Issue #72, opened 2017-07-31
HeaderMapKey implementation for String/str panics


seanmonstar:
The docs show using strings as keys, and with it being the most "ergonomic" way, people will probably default to using it everywhere. Unfortunately, a string is not always a valid header name, and because of this, the `HeaderMapKey` implementation panics if it is invalid. 

This could lead to people uniformly just using strings, and then being surprised and upset when their server panics because of an illegal string. Of course, it's better to have panicked than to allow the damaging header name through, but users certainly would not have expected it. Should we remove the impl for `String`/`str`, so that users realize that `HeaderName::try_from(str).unwrap()` can panic?

alexcrichton:
One possible alternative which may work is to only have `HeaderMapKey for &'static str` in the sense that it requires a string literal (albeit not the *only* source of `&'static str). It truly is a programmer error to have an invalid string literal as a header map key!

seanmonstar:
That sounds nice to me! If we do that, is the error message at all obvious when a user doesn't supply a static string?

carllerche:
This PR is replaced by #86?

seanmonstar:
Closed by #86.

PR #84, opened 2017-08-01
implement HeaderMapKey only for static strings

seanmonstar:
This is a possible solution for #72. This makes it so you can only use `&'static str` (besides `HeaderName`) as a key, and thus any panics are because of static strings with illegal characters. This much more closely matches the idea of panics should be for programmer errors.

**DO NOT MERGE YET.** I'm going to open another PR showing a very different solution that removes the panics completely, and we can compare.

### DIFF
--- a/src/header/map.rs
+++ b/src/header/map.rs
@@ -652,7 +652,7 @@ impl<T> HeaderMap<T> {
     /// map.append("x-hello", "world");
     /// assert_eq!(map.get("x-hello").unwrap(), &"hello");
     /// ```
-    pub fn get<K: ?Sized>(&self, key: &K) -> Option<&T>
+    pub fn get<K>(&self, key: K) -> Option<&T>
         where K: HeaderMapKey
     {
         match key.find(self) {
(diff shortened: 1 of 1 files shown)


carllerche:
As mentioned in IRC, the point of implementing `HeaderMapKey` on more types is to enable the ability to avoid the `K` -> `HeaderName` conversion when the key already exists in the map.

alexcrichton:
@carllerche just to make sure I understand, this means that a lookup which previously is:

```rust
let a = String::from("...");
foo.insert(&a, ...);
```

Previously we won't allocate a `HeaderName` because it's already in the map, but after this we're forced to look up with a `HeaderName` or a `&HeaderName`, where creating one of those from a `String` forces an allocation?

seanmonstar:
@alexcrichton for insert, it doesn't really matter. It must make an owned `HeaderName` to store as a key in the map. However, searching for keys shouldn't really need to parse and allocate a `HeaderName`.

In this PR, that would be required.

```rust
// before
map.get(&some_custom_name)

// after
let name = some_custom_name.parse().unwrap();
map.get(&name)
```

#86 explores a different solution. Since searching just requires a hash and then a comparison, `"evil\r\nname"` should never be found in the map, so `map.get("evil\r\nname").is_none()` should always be true, and not need to panic.

carllerche:
It definetly does matter for insertion as @alexcrichton pointed out. Currently, If the map already contains the key, the example will not allocate. With this change, it would need to as the HeaderName needs to be constructed before passing to insert.

carllerche:
I think this PR has been superseded by #86?

seanmonstar:
#86 is the better solution, closing.

PR #86, opened 2017-08-01
Remove all implicit panics from using a string key in HeaderMap

seanmonstar:
- Splits `HeaderMapKey` into `IntoHeaderName` and `AsHeaderName`.
- `IntoHeaderName` is a bounds on `insert` and `append`. It is only
  implemented for `HeaderName` and `&HeaderName`, since those are
  infallible conversions, so `insert`/`append` won't panic, nor do they
  need to change to return a `Result`.
- `AsHeaderName` is the bounds on `get`, `get_mut`, `contains_key`, etc.
  Any method that is just used for searching. The trait is implemented
  for `HeaderName` and for `str` variants, but they no longer panic on
  invalid string names. It simply returns `None` as the illegal header
  clearly is not in the map.
- `entry` has been changed because of this. The reason to use `entry` is
  to prevent unnecessary work if the key already exists in the map. So,
  it has a `AsHeaderName` bounds, to allow the searching to use a `&str`
  without having to allocate a `HeaderName`. If not found, and the string is
  valid, then it can allocate and insert. Because the parsing can fail,
  `entry` now returns a `Result<Entry<T>, InvalidHeaderName>`.

This also makes a possible controversial change: it removes public
export of these traits. The reason: We don't want anyone to actually
name this trait. We want to able to change the trait around as needed.
This can include the name of the traits.

The downside to not exporting them is that rustdoc won't document it,
and so people who really want to dig in the docs won't be able to find
what types concretely implement these traits.

The downside can be alleviated by updating the docs otherwise, or with a
change in rustdoc.

-----

This is a different attempt of fixing #72. The first version can be seen at #84.

### DIFF
--- a/src/header/map.rs
+++ b/src/header/map.rs
@@ -1,11 +1,13 @@
-use super::name::{HeaderName, HdrName};
-use self::sealed::Sealed;
+use super::name::{HeaderName, HdrName, InvalidHeaderName};
 
 use std::{fmt, mem, ops, ptr, vec};
 use std::collections::hash_map::RandomState;
 use std::hash::{BuildHasher, Hasher, Hash};
 use std::iter::FromIterator;
 use std::marker::PhantomData;
+
+pub use self::as_header_name::AsHeaderName;
+pub use self::into_header_name::IntoHeaderName;
 
 /// A set of HTTP headers
 ///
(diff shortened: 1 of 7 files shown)


seanmonstar:
I've updated this still allow `&HeaderName` for `insert` and `append`, and other changes. I've updated the original comment to better reflect all the changes.

carllerche:
To be a bit more explicit, when inserting a header using `&str`, `insert` will always require a conversion/allocation even if the map already contains the key. To avoid this conversion / allocation, the `entry` API must be used.

(still reviewing).

carllerche:
Another impact of this change would be that `HeaderMap` is no longer a drop-in replacement for `HashMap<HeaderName, T>`. Up until now, `HeaderMap` has been a superset of the `HashMap` API. This adds a deviation (`entry(...) -> Result<...>`).

carllerche:
Should `AsHeaderName` be implemented for `&'static str`? We've previously accepted panics when supplying static values to a conversion. It would also be a shame to lose the ergonomics of being able to use static strings...

I also think the docs for `AsHeaderName` and `IntoHeaderName` should be visible (aka, the traits should be exported). `Sealed` prevents external implementations. Otherwise, there is no way to see what types you can pass in as a key for `HeaderMap`.

seanmonstar:
> Should `AsHeaderName` be implemented for `&'static str`?

`AsHeaderName`  was implemented for `&'a str`, as it that trait is used for just lookups.

I had felt earlier as well that `IntoHeaderName` for `&'static str` seemed fine, since the same reasoning as `HeaderValue::from_static` applies: it is programmer error to use a static string that is an invalid header name, and so fine to panic. I've added that impl in.

> I also think the docs for AsHeaderName and IntoHeaderName should be visible.

I've done another shuffle so that these traits are publicly exported as `http::header::{AsHeaderName, IntoHeaderName}`, but their methods are on internal `Sealed` traits, so the function signatures aren't exposed.

seanmonstar:
I just pushed a change to change all `&K where K: AsHeaderName` to `K`. This follows the pattern of std, where arguments that are `AsRef<Something>` as taken by value. Now things like this are both possible: `map[HOST]` and `map.get("host")`.

alexcrichton:
This all sounds great to me! I personally think it's ok to deviate from `HashMap`, that's why it's a separate type!

carllerche:
I fixed the merge conflicts.

PR #93, opened 2017-08-03
Bring back Refresh standard header

carllerche:
H2 includes this header in the hpack static table.

### DIFF
--- a/src/header/mod.rs
+++ b/src/header/mod.rs
@@ -159,6 +159,7 @@ pub use self::name::{
     RANGE,
     REFERER,
     REFERRER_POLICY,
+    REFRESH,
     RETRY_AFTER,
     SERVER,
     SET_COOKIE,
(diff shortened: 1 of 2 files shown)


seanmonstar:
I see wanting to optimize for the hpack table, and I'm fine with that. 

I do think this header is a mistake in general. It has weird semantics if received in a response to a resource loaded by a page. It's also not standardized.

What do you think of having internal support, but not being public? Or at the least, docs saying why you probably should not use it?

carllerche:
IMO, just like user info in URI, we shouldn't be deviating from common usage of HTTP because we don't think it is a good idea.

Currently `Refresh` is featured in the HTTP/2.0 spec in the static table, so IMO we should just go w/ it.

Because this is a revert of the bit of #67 that broke the H2 implementation, I'm going to merge it as is. However, if you want to submit a PR w/ what you think it should be, we can discuss that.

PR #95, opened 2017-08-03
Impl HttpTryFrom<Parts> for Uri

carllerche:
This allows `Parts` to be used with the builder fns.

### DIFF
--- a/src/uri.rs
+++ b/src/uri.rs
@@ -637,6 +637,15 @@ impl From<Parts> for Uri {
     }
 }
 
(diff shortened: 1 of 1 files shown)


carllerche:
There currently is an `Into` implementation, so if there is it would be incorrect...

That said, it might be possible, so that should be audited. I will open an issue to check.

carllerche:
I created #96 to revisit the `Into` implementation. If there are error cases, they should be reflected in the `HttpTryInto` impl as part of resolving #96.

Issue #98, opened 2017-08-04
Make all "try from" functions consistently named


seanmonstar:
For instance, there is:

- `Method::from_bytes`
- `HeaderName::from_bytes`
- `StatusCode::from_u16`

vs

- `HeaderValue::try_from_bytes`
- `Uri::try_from_shared`

JJayet:
@seanmonstar Hi, I've started working on this.

I initially wanted to prefix `try_` every method which returns a Result<T, E> but looking at [Rust source files](https://github.com/rust-lang/rust/blob/ddaab61101e48f0a58ce176b5a6aff11ae36c986/src/libcore/str/mod.rs#L98), it would be inconsistant so I'm now inching the other way around.

What do you think? Should we remove every `try_`?

seanmonstar:
Yea, I filed this as an issue to allow discussion on exactly *how* we should be consistent.

PR #99, opened 2017-08-07
Fixed try_

JJayet:
Should fix https://github.com/carllerche/http/issues/98

I made two choices:
- Every public method is NOT prefix with `try_` in order to be more in line with other similar Rust implementation 
- I let the `try_`for every implementation of `HttpTryFrom` because it wants to mirror `TryFrom` which itself has a method named `try_from`

If it's not enough, just let me know.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 target
 Cargo.lock
+.idea
--- a/src/header/value.rs
+++ b/src/header/value.rs
@@ -80,7 +80,7 @@ impl HeaderValue {
     ///
     /// If the argument contains invalid header value characters, an error is
     /// returned. Only visible ASCII characters (32-127) are permitted. Use
-    /// `try_from_bytes` to create a `HeaderValue` that includes opaque octets
+    /// `from_bytes` to create a `HeaderValue` that includes opaque octets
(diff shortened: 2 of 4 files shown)


carllerche:
Idiomatic rust fns that are `from_*` do not fail. The `try_` prefix is used as a temporary measure until `TryFrom` in std becomes stable. At that point, these `try_*` fns may be deprecated.

seanmonstar:
This PR was opened because I made an issue about some methods start with try, others don't. See #98.

carllerche:
Ah, well it is referenced there. If we decide to go this route, we can reopen the PR.

alexcrichton:
I would personally advocate for this route, I don't think that we have a clear convention about `from_*` functions. Functions like `str::from_utf8` or `String::from_utf8` exist. I think the best rule of thumb is that if you call these functions a lot you don't want to keep getting reminded about "try" because you're already getting a `Result`. 

One thing I've seen though is that typically in Rust if there's a `try_foo` there's also a `foo` which is an infallible or otherwise different variant, but that doesn't seem to be the case here so I'd be fine with removing the `try_` prefix from these methods.

PR #100, opened 2017-08-07
Remove try!

JJayet:
Everything is said in the title.

As said [here](https://doc.rust-lang.org/std/macro.try.html):

> Prefer using ? syntax to try!. ? is built in to the language and is more succinct than try!. It is the standard method for error propagation.

So I removed the `try!` macro and replaced it with `?`.
Don't hesitate to close if you don't think it's worthy of a merge :)

### DIFF
--- a/src/header/value.rs
+++ b/src/header/value.rs
@@ -346,9 +346,9 @@ impl<'a> fmt::Debug for EscapeBytes<'a> {
         for &b in self.0 {
             if is_visible_ascii(b) {
                 let ch = unsafe { char::from_u32_unchecked(b as u32) };
-                try!(write!(f, "{}", ch));
+                write!(f, "{}", ch)?;
             } else {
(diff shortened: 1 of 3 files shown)


carllerche:
Switching from `try!` -> `?` is good, but it seems like the PR is not based on master? Would you be able to rebase against master and remove `6acdf772fdbb96c882dd6a9c4c63351a91d39b49?

thanks!

JJayet:
Oops, my bad, I'll do it right now :)

JJayet:
Due to me being stupid, I'll close and reopen.

seanmonstar:
(For the future) You can just `git rebase upstream/master` and then `git push -f origin branchName`, and the PR will be updated.

JJayet:
Thanks for the tip!
It's already reopened but I'll keep that in mind.

PR #101, opened 2017-08-07
Replaces all occurences of try! with ?

JJayet:
Everything is said in the title.

As said [here](https://doc.rust-lang.org/std/macro.try.html):

>Prefer using ? syntax to try!. ? is built in to the language and is more succinct than try!. It is the standard method for error propagation.

So I removed the try! macro and replaced it with ?.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 target
 Cargo.lock
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 target
 Cargo.lock
-.idea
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 target
 Cargo.lock
+.idea
```
carllerche: Could you remove this?

JJayet: @carllerche Would you be open to leave it ?

I put it there because I used [IntelliJ and the Rust plugin] (https://intellij-rust.github.io) and I figured that others might do it too.

carllerche: Those types of ignores should go in the global git ignore file.

JJayet: Got it, will fix it soon

carllerche:
Hmm... something weird is happening w/ the PR. It has a bunch of extra commits from master. It seems like it might not be based on master?

JJayet:
@carllerche Done, I don't know what went wrong but it's fixed now.

carllerche:
Thanks

PR #102, opened 2017-08-07
add HeaderValue::from_shared_unchecked

seanmonstar:
The reason for wanting this is to prevent double parsing a header value. For example, in hyper, the bytes are parsed using httparse, which already verifies the header values do not contain illegal values.

There's a technicality issue here: it's not actually `unsafe`, according to Rust's strict definition of `unsafe` meaning memory unsafety. These bytes are never converted into a `str` without validating them first.

At the same time, it is totally unsafe in a HTTP context...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,3 +30,7 @@ serde_json = "1.0"
 [[bench]]
 name = "header_map"
 path = "benches/header_map/mod.rs"
+
+[[bench]]
+name = "header_value"
+path = "benches/header_value.rs"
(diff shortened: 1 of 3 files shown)


carllerche:
I assume that it is backwards compatible to remove the `unsafe` annotation in the future?

seanmonstar:
Removing it would cause any usage to trigger the `unused_unsafe` warning. That will fail locally if you have `deny(warnings)`, but lints will never fail a build of a dependency.

carllerche:
Right, I think "backwards compatible" means that new warnings are permitted.

PR #108, opened 2017-08-10
adjust HeaderValue Debug output

seanmonstar:
- output just the value, not a struct
- wrap in quotes
- don't output value if sensitive flag is set

### DIFF
--- a/src/header/value.rs
+++ b/src/header/value.rs
@@ -1,6 +1,6 @@
 use bytes::Bytes;
 
-use std::{char, cmp, fmt, str};
+use std::{cmp, fmt, str};
 use std::error::Error;
 use std::str::FromStr;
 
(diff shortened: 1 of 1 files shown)


src/header/value.rs
```diff
@@ -289,10 +289,33 @@ impl AsRef<[u8]> for HeaderValue {
 
 impl fmt::Debug for HeaderValue {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        f.debug_struct("HeaderValue")
-            .field("value", &EscapeBytes(self.as_ref()))
-            .field("is_sensitive", &self.is_sensitive)
-            .finish()
+        if self.is_sensitive {
+            f.write_str("Sensitive")
```
alexcrichton: I wonder if we could maybe do something like:

```rust
f.debug_struct("HeaderValue")
    .field("value", &"...")
    .field("is_sensitive", is_sensitive)
    .finish()
```

just to make diffs between debugs of different `HeaderValue` instances a little more consistent perhaps?

seanmonstar: Specifically for the `if self.is_sensitive` branch?

alexcrichton: That's what I was thinking yeah, basically just having both the sensitive and non-sensitive versions have roughly the same shape

seanmonstar: Ah, this PR removed outputting a struct at all. So, normal output would look like this:

```
{"transfer-encoding": "chunked"}
```

And sensitive looks like this:

```
{"authorization": Sensitive}
```

alexcrichton: Ack sorry I was misleading myself! You can ignore m!

PR #109, opened 2017-08-11
convert all inherent try_from_* functions to from_*

### LINKED ISSUE #98: Make all "try from" functions consistently named

seanmonstar:
~~Some places use `from_u16`, others use `try_from_bytes`, this PR unifies the pattern. It settled on `try_from_*` for these reasons:~~

- ~~There exists a few functions, like `from_static`, which don't return a `Result`.~~
- ~~While there are several cases in std, like `from_utf8`, `from_str`, it's also inconsistent within std, with things like `Duration::from_secs() -> Duration`.~~
- ~~It lines up closer with the eventual `TryFrom::try_from`.~~

Updated to do the reverse.

Closes #98

### DIFF
--- a/src/header/value.rs
+++ b/src/header/value.rs
@@ -80,7 +80,7 @@ impl HeaderValue {
     ///
     /// If the argument contains invalid header value characters, an error is
     /// returned. Only visible ASCII characters (32-127) are permitted. Use
-    /// `try_from_bytes` to create a `HeaderValue` that includes opaque octets
+    /// `from_bytes` to create a `HeaderValue` that includes opaque octets
     /// (128-255).
     ///
     /// This function is intended to be replaced in the future by a `TryFrom`
@@ -90,7 +90,7 @@ impl HeaderValue {
     ///
     /// ```
     /// # use http::header::HeaderValue;
-    /// let val = HeaderValue::try_from_str("hello").unwrap();
+    /// let val = HeaderValue::from_str("hello").unwrap();
     /// assert_eq!(val, "hello");
     /// ```
     ///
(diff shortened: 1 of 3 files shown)


alexcrichton:
I personally disagree with the `try_*` prefix as outlined in [this comment](https://github.com/carllerche/http/pull/99#issuecomment-320812559). I don't think there's strong precedent for requiring this and I don't think the presence of infallible `from_*` functions should make all infallible functions *worse* to call. It's already going to be pretty obvious what's fallible and what's not when reading source, it's whether or not there's a `?` at the end.

I'd still value consistency more, though, so if others don't agree then I'm fine landing this.

seanmonstar:
I wasn't sure which way I leaned until I started this PR, when I saw `from_static`. That the compiler will remind you the function returns a `Result` is actually a fair reason to not have `try_` everywhere, in my opinion.

Anyone else?

carllerche:
Given the pattern of having `From<Foo>` implementations also have a struct fn `from_foo`, I opted for `TryFrom<Foo>` implementations to mirror this as `try_from_foo`.

The only question is if a type will ever implement **both** `From<Foo>` and `TryFrom<Foo>`. Because, if this happens, there could be a naming conflict.

seanmonstar:
In the RFC for `TryFrom` it's suggest that any `impl From<Foo> for Bar` would fit into a blanket `impl TryFrom<Foo> for Bar { type Error = ! }`.

carllerche:
If we are confident that there is never any naming conflict, then it seems fine to make the style change as @alexcrichton suggested.

seanmonstar:
I've updated to do the reverse, then. All inherent `try_from_*` becomes `from_*`.

PR #113, opened 2017-08-16
HeaderMap::get_all() returns a GetAll instead of an Option

### LINKED ISSUE #111: HeaderMap::get_all should not return Option
Currently, `HeaderMap::get_all` returns an `Option<GetAll<T>>`, with `None` indicating the absence of any header with the specified name. That type signature allows encoding "impossible" situations, which code will then have to handle. Specifically, it allows `Some` of an empty iterator. That's going to force code using this type to have extra logic dismissing that "impossible" case.

Please switch the type to just `GetAll<T>`, with an empty iterator indicating the absence of any header with the specified name. It's still trivial to detect presence/absence of a header (such as via `contains_key`), and this will simplify callers, reducing uses of `.unwrap()` and similar.

seanmonstar:
Also removes `key()` and `get()` from `GetAll`.

Closes #111

### DIFF
--- a/src/header/map.rs
+++ b/src/header/map.rs
@@ -138,7 +138,7 @@ pub struct Drain<'a, T> {
 #[derive(Debug)]
 pub struct GetAll<'a, T: 'a> {
     map: &'a HeaderMap<T>,
-    index: usize,
+    index: Option<usize>,
 }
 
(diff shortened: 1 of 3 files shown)


alexcrichton:
👍

PR #120, opened 2017-08-25
Add HeaderName::from_lowercase

carllerche:
HTTP/2.0 requires that all header names be provided with only lowercase
characters. Using uppercase characters must result in a PROTOCOL_ERROR.
In order to avoid scanning the string multiple times while parsing,
provide a lowercase parse variant.


carllerche:
Waiting on @seanmonstar. I believe he had reservations.

PR #123, opened 2017-09-06
added shortcut methods to create request builder with method and uri

krampenschiesser:
I am using the http library in my project and I was missing shortcut methods to create requests in a simple manner.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -212,6 +212,196 @@ impl Request<()> {
     pub fn builder() -> Builder {
         Builder::new()
     }
+
+
+    /// Creates a new `Builder` initialized with a GET method and the given URI.
+    ///
+    /// This method returns an instance of `Builder` which can be used to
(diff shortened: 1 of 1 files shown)


carllerche:
The idea is good, I wonder if these fns should return requests directly instead of builders.

Also, we know that POST and PUT requests will have bodies, so should the body be accepted as an argument?

krampenschiesser:
I would prefer builders, in order to add occasional headers or, as you said, a body.

seanmonstar:
> Also, we know that POST and PUT requests will have bodies, so should the body be accepted as an argument?

POST may have a body, but lots of POST endpoints with an empty body also exist. It is pretty weird to have PUT without a body... but it's also super easy to add a body after.

carllerche:
I'll allow it.

carllerche:
Thanks!

PR #125, opened 2017-09-08
Update crate metadata

carllerche:


seanmonstar:
This metadata was added when preparing the `v0.1.0` tag.

PR #129, opened 2017-09-26
Provide Uri accessor for Authority.

carllerche:
Currently, there is no way to access the `Authority` component of `Uri`
at a type level. This is desireable because:

* `Authority` is backed by `Bytes` (for clone).
* `Authority` should be case-insensitive for Eq / Hash.

Unfortunately, `Uri::authority` already exists and returns a `&str`.
This is an oversight that originated from the original iteration of
`Uri` not providing typed components.

This PR adds an `authority_part` accessor to URI and deprecates
`authority` with the expectation that the next semver release of `http`
will rename `authority_part` back to `authority`.

This PR also provides an `Eq` and `Hash` implementation for `Authority`
and adds `Authority::port`.

### DIFF
--- a/src/uri.rs
+++ b/src/uri.rs
@@ -456,6 +456,8 @@ impl Uri {
     ///                 authority
     /// ```
     ///
+    /// This function will be renamed to `authority` in the next semver release.
+    ///
     /// # Examples
     ///
     /// Absolute URI
@@ -464,7 +466,7 @@ impl Uri {
     /// # use http::Uri;
     /// let uri: Uri = "http://example.org:80/hello/world".parse().unwrap();
     ///
-    /// assert_eq!(uri.authority(), Some("example.org:80"));
+    /// assert_eq!(uri.authority_part().map(|a| a.as_str()), Some("example.org:80"));
     /// ```
     ///
     ///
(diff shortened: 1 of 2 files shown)


src/uri.rs
```diff
@@ -1765,7 +1870,7 @@ test_parse! {
     [],
 
     scheme = Some("http"),
-    authority = Some("127.0.0.1"),
+    authority_part = Some(&"127.0.0.1".parse().unwrap()),
```
seanmonstar: All of these seem unfortunate. Can we adjust the macro to allow us to still just include a string in the test cases?

src/uri.rs
```diff
@@ -1033,6 +1082,62 @@ impl Authority {
...
+/// let mut s = DefaultHasher::new();
+/// b.hash(&mut s);
+/// let b = s.finish();
+///
+/// assert_eq!(a, b);
+/// ```
+impl Hash for Authority {
+    fn hash<H>(&self, state: &mut H) where H: Hasher {
+        self.data.as_bytes()
+            .to_ascii_lowercase().hash(state);
```
seanmonstar: This creates a new `Vec` and copies into it, right? Could the hash instead do a loop with each byte calling `b.to_ascii_lowercase()` instead?

carllerche: Ah, I think I ended up copying this from somewhere.

src/uri.rs
carllerche: I didn't see an obvious way to adjust the macro.

PR #135, opened 2017-10-30
Parse percentage encoded paths

### LINKED ISSUE #134: Can not parse uri
can not parse uris `/echo/abcdefgh_i-j%20/abcdefg_i-j%20478` or `/echo/abcdefgh_i-j /abcdefg_i-j 478`

carllerche:
Fixes #134

### DIFF
--- a/src/uri.rs
+++ b/src/uri.rs
@@ -229,6 +229,36 @@ const SCHEME_CHARS: [u8; 256] = [
         0,     0,     0,     0,     0,     0                              // 25x
 ];
 
+const HEX_DIGIT: [u8; 256] = [
+    //  0      1      2      3      4      5      6      7      8      9
+        0,     0,     0,     0,     0,     0,     0,     0,     0,     0, //   x
+        0,     0,     0,     0,     0,     0,     0,     0,     0,     0, //  1x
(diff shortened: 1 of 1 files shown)


src/uri.rs
```diff
@@ -1200,12 +1230,30 @@ impl PathAndQuery {
...
-        for i in 0..src.len() {
+        let mut i = 0;
+
+        while i < src.len() {
             let b = src[i];
 
             match URI_CHARS[b as usize] {
                 0 => {
-                    return Err(ErrorKind::InvalidUriChar.into());
+                    if b == b'%' {
```
seanmonstar: Could this just follow the pattern of `?` and `#` instead of matching 0 and then checking again what byte it is?

carllerche: `URI_CHARS` was also used in other situations where `%` isn't permitted. It seemed roughly equivalent to do it this way vs. create `PATH_CHARS` that permitted `%`.

seanmonstar: Oh, OK!

nox:
Which spec is this patch following? It seems that now Hyper cannot cope with a URL containing an unencoded `|` char in its query.

PR #140, opened 2017-11-09
add Uri::scheme_part, deprecate Uri::scheme

### LINKED ISSUE #132: Improve `uri::Scheme` so that it can be used better as its own type.
This issue represents applying the same work that `Authority` received in https://github.com/hyperium/http/commit/05ec166c4fe586ae22c7d7b9c0958022f4c8bbe4 to `Scheme`.

seanmonstar:
Closes #132


carllerche:
Thanks for keeping the file re-org in a separate commit. I will want to merge this PR keeping these two commits.

Could we move the uri tests -> `tests/uri.rs` instead? That should help speed up compiling a bit when the tests are modified.

Looks good to me though.

seanmonstar:
Looks like there a usage of `ErrorKind` in there, which is private. We could move all the other tests... Though, FWIW, being in a separate module should mean things are still quick when using incremental compilation.

PR #150, opened 2017-12-12
Implement PartialOrd/PartialEq for Authority and PathAndQuery

### LINKED ISSUE #149: Implement PartialEq / PartialOrd for PathAndQuery
PathAndQuery should be comparable against strings. It should probably also include an as_str() fn and a number of other fns that I am forgetting about.

kbacha:
Implementation of PartialOrd and PartialEq for Authority and PathAndQuery. Implementations are bidirectional and units tests are added for all. Authority has been implemented with case-insensitivity in mind.

cc: @carllerche 

closes #149

### DIFF
--- a/src/uri/authority.rs
+++ b/src/uri/authority.rs
@@ -1,6 +1,6 @@
 #[allow(unused)]
 use std::ascii::AsciiExt;
-use std::fmt;
+use std::{cmp, fmt, str};
 use std::hash::{Hash, Hasher};
 use std::str::FromStr;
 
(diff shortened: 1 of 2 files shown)


carllerche:
Yeah, this looks correct to me :)

kbacha:
@carllerche Sweet, I've wrapped up the implementation on Authority and PathAndQuery

src/uri/authority.rs
```diff
@@ -192,6 +192,64 @@ impl PartialEq<str> for Authority {
...
+
+/// Case-insensitive ordering
+///
+/// # Examples
+///
+/// ```
+/// # use http::uri::Authority;
+/// let authority: Authority = "DEF.com".parse().unwrap();
+/// assert!(authority < *"ghi.com");
+/// assert!(authority > *"abc.com");
```
seanmonstar: Curious, is the deref needed? Is that because there is an impl for `str`, but not `&'a str`?

src/uri/authority.rs
```diff
@@ -192,6 +192,64 @@ impl PartialEq<str> for Authority {
...
+/// ```
+impl PartialOrd for Authority {
+    fn partial_cmp(&self, other: &Authority) -> Option<cmp::Ordering> {
+        self.data.to_ascii_lowercase().partial_cmp(&other.data.to_ascii_lowercase())
```
seanmonstar: All of these comparisons actually allocate and copy a new string before comparing. It's probably faster to just compare the bytes individually. Something like:

```rust
let left = self.data.iter().map(|b| b.to_ascii_lowercase());
let right = other.data.iter().map(|b| b.to_ascii_lowercase());
Some(left.cmp(right))
```

src/uri/path.rs
```diff
@@ -256,6 +286,80 @@ impl fmt::Display for PathAndQuery {
     }
 }
 
+// ===== PartialEq / PartialOrd =====
+
+impl PartialEq for PathAndQuery {
+    #[inline]
+    fn eq(&self, other: &PathAndQuery) -> bool {
+        self.as_str() == other.as_str()
```
seanmonstar: Does `ByteStr` not have a `PartialEq` implementation that can just be used? Would let us skip the `is_empty` checks in both `as_str` calls.

src/uri/authority.rs
kbacha: I copied the other doc test here for `PartialEq`
I did find that either the str needs to deref or authority need to be a ref. Perhaps there's a better way? (I'm a little new to rust so I'm still working through some of these details.)

src/uri/path.rs
kbacha: Aw, probably does. Would that just be `self.data == other.data`?

src/uri/authority.rs
kbacha: Nice, I'll give it a try.

kbacha: @seanmonstar I'm having issue with calling `.iter()` on the data (which I'll try to figure out).

I'm looking at the implementation of `to_ascii_lowercase` and it creates a vec and returns back (unsafe) a string. Since this proposal would still map all the bytes to another vec, wouldn't the cost be comparable?

src/uri/authority.rs
seanmonstar: I believe if you add an implementation of `PartialEq<&'a str> for Authority` and `PartialEq<Authority> for &'a str`, then it can be written without the derefs.

src/uri/authority.rs
seanmonstar: Oh, my bad, try `data.as_bytes().iter()`.

The difference is that the `to_ascii_lowercase` will reserve space in the heap, and then copy the bytes into that, before doing the comparisons, so doing it this way actually can be rather noticeable.

src/uri/path.rs
seanmonstar: Hm, is it possible to end up with that case?

But yea, looks like `ByteStr` does implement `PartialEq`, so it could just be `==`. If we can use that, it could take advantage of pointer comparisons if two `Authority`s were backed by the same `Bytes` buffer.

src/uri/authority.rs
kbacha: oh, cool, that worked!

kbacha:
@seanmonstar How are these changes? a60fd89

PR #151, opened 2017-12-22
Add a `map` to Request and Response

### LINKED ISSUE #143: Provide `map` for bodies
Right now, there's no easy way to swap out the body while changing its type (i.e., a `map` method). That's very useful when transforming requests and responses.

kbacha:
For ease of use, it's nice to be able to map using a closure from
one type to another. This adds that ability to both the
`Request` and `Response` objects.

closes #143


src/request.rs
```diff
@@ -638,6 +638,27 @@ impl<T> Request<T> {
...
+    /// # use http::*;
+    /// let request = Request::builder().body("some string").unwrap();
+    /// let mapped_request: Request<&[u8]> = request.map(|b| {
+    ///   assert_eq!(b, "some string");
+    ///   b.as_bytes()
+    /// });
+    /// assert_eq!(mapped_request.body(), &"some string".as_bytes());
+    /// ```
+    #[inline]
+    pub fn map<O, F>(self, op: F) -> Request<O>
```
carllerche: Super nitty, but Rust convention seems to be naming the function argument of a functional combinator `f` (and this will show up in the rustdoc).

Could you rename `op` -> `f`. Also, could you rename `O` -> `U`.

(I am mostly going by observations that I have made of std).

kbacha: Yes of course! I actually had it like that originally but ended up changing it because the `Result` struct in std uses `op`:
https://doc.rust-lang.org/std/result/enum.Result.html#method.map

kbacha: I'm also going to swap `U` and `F` as type arguments `map<F, U>` since the function type appears first in the signature.

PR #153, opened 2017-12-24
Implements PartielEq for StatusCode with u16

### LINKED ISSUE #152: Impl `PartialEq` and `PartialOrd` on StatusCode for u16
We can convert to/from u16, it would be nice to just be able to compare status codes to u16.

kbacha:
As a convenience, it's shorter to write:

`status == 400`

rather than:

`status.as_u16() == 400`
or
`status == StatusCode(400)`

I was writing tests for builder convenience methods with status and realized that it might be simpler to be able to compare a status code directly with a u16 rather than needing to call a method. So I inlined the implementations of eq.

closes #152

### DIFF
--- a/src/status.rs
+++ b/src/status.rs
@@ -170,6 +170,20 @@ impl Default for StatusCode {
     }
 }
 
+impl PartialEq<u16> for StatusCode {
+    #[inline]
+    fn eq(&self, other: &u16) -> bool {
+        self.as_u16() == *other
+    }
+}
+
+impl PartialEq<StatusCode> for u16 {
+    #[inline]
+    fn eq(&self, other: &StatusCode) -> bool {
+        *self == other.as_u16()
+    }
+}
+
 impl From<StatusCode> for u16 {
     #[inline]
     fn from(status: StatusCode) -> u16 {
--- a/tests/status_code.rs
+++ b/tests/status_code.rs
@@ -13,6 +13,13 @@ fn from_bytes() {
     }
 }
 
+#[test]
+fn equates_with_u16() {
+    let status = StatusCode::from_u16(200u16).unwrap();
+    assert_eq!(200u16, status);
+    assert_eq!(status, 200u16);
+}
+
 macro_rules! test_round_trip {
     ($($num:expr,)+) => {
         #[test]


carllerche:
Could you elaborate on the use case where you want to do `status < 400` instead of `!status.is_client_error()`?

Perhaps, what you are hoping for is a fn to check if the status code is an error or not? Basically:

```rust
fn is_error(&self) -> bool {
    self.is_client_error() || self.is_server_error()
}
```

kbacha:
@carllerche Yup, that's much of the time what I use it for! With that reframing of the problem I can think of several useful ones and it looks like they have mostly been implemented.

Slightly more efficient if we implement just those comparisons right? (maybe llvm would optimize for us?)

```rust
fn is_error(&self) -> bool {
    600 >self.0 && self.0 >= 400
}
```

I suppose I was thinking it might be easier to have the u16 shorthand instead of writing either:
```ruby
assert!(status.as_u16() == 201);
assert!(status == StatusCode(201);
```

But I think that's what the constants might be for anyhow. It just came up while I was working on another issue, if there's no value in it, we can close it.

carllerche:
I'm not necessarily against adding comparisons to `u16`. `Eq` and `PartialEq` are good additional implementations I think. I just want to be a bit cautious before adding `Ord` as I think the `is_*` tests may be more appropriate.

carllerche:
Also, `http` doesn't permit creating status codes above 599. So we only need to check if it is 4xx or 5xx.

kbacha:
@carllerche Ah, yea, I can totally remove the `Ord` implementation.

kbacha:
@carllerche updated without Ord.

seanmonstar:
I'm torn.

- It certainly is easier to be able to just compare with numbers. Most people understand `if res.status() == 200` just fine...
- However, I do think that it can make things more obscure, especially if dealing with less common status codes. For instance, compare `if status == 431` vs `if status == REQUEST_HEADERS_TOO_LARGE`. It's pretty clear what case you're looking for in the latter instance.

I think the second way, while slightly painful when writing, is much better when Future You or Future Colleague comes around and has to *read* it. In that sense, I might be trying to push this opinion into a basic library, but then, that might not be so bad...

kbacha:
If I'm understanding this would mean something like for non-standard status codes.

```rust
const ENHANCE_YOUR_CALM: http::StatusCode = http::StatusCode(420);

// ...

if res.status() == ENHANCE_YOUR_CALM {
}
```

And if I use my own status codes I might still just write something like:

```rust
if res.as_u16() == 420 {
}
```

or 

```rust
if res == http::StatusCode(420) {
}
```

I'm not sure that this opinion would necessarily solve anything if people use a custom status code (like 420), unless `http` enventually migrates the status codes from u16 to an enum.

Having said that, this is trivial and I'm cool with closing.

PR #154, opened 2017-12-25
Implement helper functions for response builder

### LINKED ISSUE #131: Implement `Response` builder helper fns
Just like `Request` has `get`, `post`, etc... helper fns, there should be equivalent helpers for `Response`.

choubacha:
Adds the following helpers to make building easier:
* `content_type`
* `http_2`
* Various common status codes like `ok` and `not_found`

I figured these would be useful and common enough to warrant special methods for them. If you can think of any more I'll add them.

closes #131

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -231,6 +231,261 @@ impl Response<()> {
     pub fn builder() -> Builder {
         Builder::new()
     }
+
+    /// Sets the status to OK (200)
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use http::*;
+    ///
+    /// let response = Response::ok()
+    ///     .body(())
+    ///     .unwrap();
+    ///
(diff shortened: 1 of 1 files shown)


carllerche:
I like the `http_2` and `content_type` helpers as is!

For the other helper fns, I was imagining that they would be used instead of calling `builder`:

```rust
let response = Response::permanently()
    .body(())
    .unwrap();
```

Mostly because all responses must have a status code AND you will only set it once (most likely). If you want to change the status code after the fact, you can always call the `status` function.

kbacha:
@carllerche Check 954e1f0 to see my changes. Is that what you were thinking?

seanmonstar:
I have a conflicting opinion (sorry!), which I've added to https://github.com/hyperium/http/issues/131#issuecomment-355373329

kbacha:
@seanmonstar Good thing I left it as a separate commit!

PR #159, opened 2018-01-05
Implement variant on version to require catch-all

### LINKED ISSUE #155: `Version` can be exhaustively matched
I feel like having `Version` be a load of `const`s rather than an `enum` was probably so that it could be extended without breakage. Unfortunately it seems `rustc` is too smart and sees through the ruse?

```rust
fn foo(version: Version) {
    match version {
        Version::HTTP_09 => http09(),
        Version::HTTP_10 => http10(),
        Version::HTTP_11 => http11(),
        Version::HTTP_2 => http2(),
        v => panic!("unsupported http version {:?}", v) // unreachable pattern lint triggered
    }
}
```

kbacha:
To require developers to implement the catch-all state of versions (for versions that are unknown to the crate) I've added a private variant.

closes #155 

Not sure how to mark this PR against 0.2 milestone.


seanmonstar:
Thanks! The hope is that there is only 1 set of breaking changes, in 0.2, as a release candidate for 1.0. So, we'll probably hold off for a little bit while we collect a list of all breaking changes that might have to happen (also likely to happen is an upgrade of the `bytes` dependency, for instance).

PR #162, opened 2018-01-12
websocket handshake related headers

fafhrd91:
WebSockets handshake related headers ([rfc6455](https://tools.ietf.org/html/rfc6455))


carllerche:
Thanks for the PR. I'm :+1: on adding these headers.

CI is failing because you also need to re-export the types here: https://github.com/fafhrd91/http/blob/master/src/header/mod.rs#L107.

fafhrd91:
oh! 
re-exported types

seanmonstar:
Cool, thanks!

PR #164, opened 2018-01-19
parsing an empty string as an Authority returns an error

### LINKED ISSUE #163: Parsing Authority with an empty string should be an error
Consider:

```rust
let auth = "".parse::<Authority>().unwrap();
```

That currently succeeds, and gives you an `Authority` that is blank. If you then try to construct a `Uri` with it, by using `uri::Parts`, and then to string it, you can end up with `http:///`, which is clearly wrong.

seanmonstar:
Closes #163

### DIFF
--- a/src/uri/authority.rs
+++ b/src/uri/authority.rs
@@ -42,7 +42,7 @@ impl Authority {
     /// # }
     /// ```
     pub fn from_shared(s: Bytes) -> Result<Self, InvalidUriBytes> {
-        let authority_end = Authority::parse(&s[..]).map_err(InvalidUriBytes)?;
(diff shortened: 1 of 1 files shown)


src/uri/authority.rs
```diff
@@ -53,6 +53,7 @@ impl Authority {
         })
     }
 
+    // Note: this may return an *empty* Authority. You might want `parse_non_empty`.
     pub(super) fn parse(s: &[u8]) -> Result<usize, InvalidUri> {
```
carllerche: This looks to be crate private. Is there a reason the caller could pass in an empty authority?

seanmonstar: Yes, there is, in `parse_full` for `Uri`. An empty authority is handled differently there (error is `MissingAuthority` instead of `Empty`).

PR #180, opened 2018-03-02
Add X-Requested-With header name

niklasf:
A non-standard but very commonly used header that can be used as a defence against CSRF.


jrozner:
For what it's worth it shouldn't be possible to forge the header, unless you're writing your own JS on the page, without existing XSS on the page already and once there is XSS on the page any CSRF protection is by-passable anyway. And if you are writing your own JS on the page you're only going to CSRF yourself so it's sort of a moot point.

niklasf:
There was the problem that Flash was able to send the header across sites. It has been fixed in Chrome and was open for a long time in Safari. Not sure about the status there.

robjtede:
4 years later...

Flash is dead. `fetch()` is used in web apps more commonly alongside proper CORS rules. `Same-Site` cookies are becoming more common which practically eliminate CSRF problems.

Doesn't seem worth it at this point to add since [it's easy enough to have const `HeaderName`s using `once_cell` already](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ca8fdddfe9db7f2694c8de67d5a48d59).

niklasf:
I agree.

PR #185, opened 2018-03-09
deps: update ordermap to 0.4

ignatenkobrain:


carllerche:
I believe that this is now [indexmap](https://crates.io/crates/indexmap). Would you mind switching to that instead?

Thanks for the work!

ignatenkobrain:
Yeah, time flies. I will update to indexmap.

ignatenkobrain:
pushed, let's see.

carllerche:
Thanks!

PR #186, opened 2018-03-09
add from_static for Authority

### LINKED ISSUE #178: Add `uri::Authority::from_static`
This would be similar to the `HeaderValue::from_static`, panicking instead of returning a `Result`.

If it's possible for the static slice to be used instead of copying it, that'd be bonus points.

rleungx:
Resolves #178.

### DIFF
--- a/src/uri/authority.rs
+++ b/src/uri/authority.rs
@@ -73,10 +73,10 @@ impl Authority {
     pub fn from_static(src: &'static str) -> Self {
         let s = src.as_bytes();
         let b = Bytes::from_static(s);
-        let authority_end = Authority::parse_non_empty(&b[..]).unwrap();
+        let authority_end = Authority::parse_non_empty(&b[..]).expect("static str is not valid authority");
 
         if authority_end != b.len() {
-            panic!("invalid uri character");
+            panic!("static str is not valid authority");
         }
 
         Authority {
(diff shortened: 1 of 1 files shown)


src/uri/authority.rs
```diff
@@ -53,6 +53,37 @@ impl Authority {
...
+    ///
+    /// ```
+    /// # use http::uri::Authority;
+    /// let authority = Authority::from_static("example.com");
+    /// assert_eq!(authority.host(), "example.com");
+    /// ```
+    pub fn from_static(src: &'static str) -> Self {
+        let s = src.as_bytes();
+        let b = Bytes::from_static(s);
+        let authority_end = Authority::parse_non_empty(&b[..]).unwrap();
```
seanmonstar: Perhaps this should be `.expect("static str is not valid authority")`, so that the panic message is clearer.

src/uri/authority.rs
```diff
@@ -53,6 +53,37 @@ impl Authority {
...
+
+        if authority_end != b.len() {
+            panic!("invalid uri character");
```
seanmonstar: This happens for things like `Authority::from_static("hyper.rs/guides")`, right? So, what should the message say? Same as the `expect` above?

Issue #168, opened 2018-02-09
Add a `HeaderName::from_static` method


seanmonstar:
This would be similar to the `HeaderValue::from_static`, panicking instead of returning a `Result`.

If it's possible for the static slice to be used instead of copying it, that'd be bonus points.

smithsps:
As all header names are represented as lower case, how should a static strings with uppercase characters be handled with regards to using the static slice vs copying? 

I couldn't decide on the best approach out of these four:

1. Panicing/Warning on uppercase. As it is already panicing on errors, it could require that all passed strings to be lowercase. This would always allow the use the static slice rather than copying, however its probably most annoying to users. 

1. Ignoring case. It can always copy and convert to lowercase as is done in HeaderName::from_bytes.

1. Strict Hybrid. Combining options 1 and 2: On header strings of less than 64 it could ignores case and always copy the string. With strings of >=64 it could panic and require all characters to be lowercase.

1. Compatibility Hybrid. Scan if the static strings have uppercase characters and then appropriately use the static slice or copying as needed.

seanmonstar:
I'd personally prefer either 1 or 4, perhaps leaning towards 1 since it's a static string. @carllerche?

carllerche:
I would go with 1: panic. This is the strategy used elsewhere. Switching to 4 at a later time if this is wrong is not a breaking change.

seanmonstar:
Done in #195

katie-martin-fastly:
:wave: Hello! Would a PR to switch to option 4 described above, copying the string when capital letters are found, be a welcome change? I have found this panic behavior confusing in the past, and don't believe that I'm alone in that experience.

I saw your comment [here](https://github.com/hyperium/http/issues/228#issuecomment-404996755) that says:

> As you've pointed out, RFC 7230 says names are case-insensitive**[...]** Additionally, in HTTP/2, all header names are lowercase (it's specified to be a protocol error for upper case names). Because of this, HeaderName chooses to always handle names as lower case. Doing this allows optimizations [that improve the] speed of inserts and searches in the HeaderMap.

I'd like to first be clear that I agree with internally lower-casing header names for these performance optimizations, have no intention of asking for that behavior to change, and thank you for maintaining such a wonderful crate here. :heart:

[...]

PR #195, opened 2018-04-03
header name: add from_static method

smithsps:
Implements: #168 : `HeaderName::from_static`

Converts a static string to a HTTP header name.
Function panics on an invalid headers, and requires lowercase characters, as so the static string will have the same format as the internal format.

Function should also be zero copy, beyond what is done in parse_hdr, with the static slice being used for the `ByteStr` in custom headers.

### DIFF
--- a/src/header/name.rs
+++ b/src/header/name.rs
@@ -2008,90 +2008,90 @@ fn eq_ignore_ascii_case(lower: &[u8], s: &[u8]) -> bool {
     })
 }
 
-#[test]
-fn test_bounds() {
-    fn check_bounds<T: Sync + Send>() {}
-    check_bounds::<HeaderName>();
-}
-
-#[test]
-fn test_parse_invalid_headers() {
-    for i in 0..128 {
-        let hdr = vec![1u8; i];
-        assert!(HeaderName::from_bytes(&hdr).is_err(), "{} invalid header chars did not fail", i);
-    }
-}
-
-#[test]
-fn test_from_hdr_name() {
-    use self::StandardHeader::Vary;
-
-    let name = HeaderName::from(HdrName {
-        inner: Repr::Standard(Vary),
-    });
-
-    assert_eq!(name.inner, Repr::Standard(Vary));
-
-    let name = HeaderName::from(HdrName {
-        inner: Repr::Custom(MaybeLower {
-            buf: b"hello-world",
-            lower: true,
-        }),
-    });
-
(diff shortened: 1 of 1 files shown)


smithsps:
Nightly build breaking on `std::ascii::AsciiExt` deprecation, which is currently fixed with the https://github.com/hyperium/http/pull/194 PR's added flags.

src/header/name.rs
```diff
@@ -1594,6 +1594,67 @@ impl HeaderName {
...
+    /// 
+    /// ```should_panic
+    /// # use http::header::*;
+    /// #
+    /// // Parsing a header that contains invalid symbols(s):
+    /// HeaderName::from_static("content{}{}length"); // This line panics!
+    /// 
+    /// // Parsing a header that contains invalid uppercase characters.
+    /// let a = HeaderName::from_static("foobar");
+    /// let b = HeaderName::from_static("FOOBAR"); // This line panics!
```
seanmonstar: It does? Looking at the implementation, it *looks* like the block with `lower: false` will still parse successfully. I might not understand all the details of this file, though...

src/header/name.rs
```diff
@@ -2024,3 +2085,67 @@ fn test_eq_hdr_name() {
...
+    // Repr::Standard cases
+    let a = HeaderName { inner: Repr::Standard(Vary) };
+    
+    let b = HeaderName::from_static("vary");
+    assert_eq!(a, b);
+
+    let b = HeaderName::from_static("vaary");
+    assert_ne!(a, b);
+
+    let result = panic::catch_unwind(|| HeaderName::from_static("Vary"));
```
seanmonstar: What if each of these that should panic were separate `#[test] #[should_panic] fn foo() {}` test cases, so that `catch_unwind` wouldn't be needed? I think it might be clearer in reading. For example:

```rust
#[test]
#[should_panic]
fn from_static_uppercase() {
    HeaderName::from_static("Vary");
}
```

Since this would make a bunch of test functions, it might also be a good idea to push these into an inner `#[cfg(test)] mod tests { ... }`. Whatcha think?

src/header/name.rs
smithsps: It is a little weird, but should be correct with the same checking method in `from_lower`. 
However, I agree it is unclear and can be improved.

src/header/name.rs
smithsps: Definitely agree, and the other tests could also be moved into the block too.

smithsps:
No worries, and thanks for all your work on Hyperium!

PR #203, opened 2018-05-24
Implement Clone for header::ToStrError

ivanovaleksey:
Hello,

Thank you for the great library!

I faced with the problem where I need `header::ToStrError` to implement `Clone` trait.
Let me describe my use case.

I am working on web service with actix-web (as transport) and jsonrpc-core (as business logic).
I need to implement authentication via JWT.
So, at transport layer I fetch `Authorization` header and convert it to `Result<String, ToStrError>` if it is present. You can see it [here](https://github.com/netology-group/iam/pull/15/files#diff-b4aea3e418ccdb71239b96952d9cddb6R79).
In order to pass it to business logic layer I need to use jsonrpc-core [`Metadata`](https://docs.rs/jsonrpc-core/8.0.1/jsonrpc_core/trait.Metadata.html) feature.
I defined my metadata as [the following](https://github.com/netology-group/iam/pull/15/files#diff-c4f869d1c823b426c320d58fb8cab701R28).
And because of `Metadata: Clone` it requires `header::ToStrError: Clone`.

There are few reasons for passing `Option<Result<String, actix_web::http::header::ToStrError>>` via metadata and not just `String`:
- It allows me to have a single point for checking token presence and signature.
- According to JSON-RPC specification I need to provide corresponding `id` on `Unauthorized` responses. For that I need to parse request first and get `id`s from there. There are different kinds of requests: single, batch, method call, notification.

Maybe this code could be written in another way, i.e. moving authentication middleware to actix-web side. But I think current approach where actix-web is only acting as transport layer is more clear and correct.

So, I thought that we could implement `Clone` for `header::ToStrError`.
I am not sure if doing this has any drawbacks.
Please, tell me what you think about that.

Thank you.

### DIFF
--- a/src/header/value.rs
+++ b/src/header/value.rs
@@ -38,7 +38,7 @@ pub struct InvalidHeaderValueBytes(InvalidHeaderValue);
 ///
 /// Header field values may contain opaque bytes, in which case it is not
 /// possible to represent the value as a string.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct ToStrError {
     _priv: (),
 }


seanmonstar:
I actually might have pushed back a little, but oh well. Mostly because it's actually very rare for error types to implement `Clone` in Rust, and if the type were to ever encapsulate another error internally, it probably would break our `Clone` impl.

In the use case in the description, I'd recommend either passing the `HeaderValue` itself, and do the parsing where it can handle the error, or map the `ToStrError` to something specific to your application.

carllerche:
Hmm, @seanmonstar brings up a good point. I’m thinking that this should be reverted in favor of forwards compat.

Does the proposed solution work in your case?

ivanovaleksey:
Thank you for the feedback!

Passing `HeaderValue` is really a good option. I think it is even more clear solution regardless `Close` implementations. Yesterday I didn't see it at all.

To be honest, I don't understand the problem with break.
> if the type were to ever encapsulate another error internally, it probably would break our Clone impl.

Could you please explain it a little bit more to me?

Feel free to revert it. I am sorry if I made inconvenience.

seanmonstar:
@ivanovaleksey:

> Could you please explain it a little bit more to me?

Sure. If we were to change the internals of the type to `struct ToStrError { inner: FromUtf8Error }`, for instance, the inner error type isn't `Clone`, so we couldn't continue to provide `Clone` for `ToStrError`. (In that specific case, we sort could, but cloning a big `String` is expensive, and it's just to give an example).

> I am sorry if I made inconvenience 

Not at all! On the contrary, we were chatting about how awesome it was that you wrote so much detail about why you wanted the change in the first place. We really appreciate it! I hope it doesn't scare you off :D

ivanovaleksey:
@seanmonstar thank you for the reply, now I see!

PR #218, opened 2018-06-20
add conversions of integers into HeaderValue

### LINKED ISSUE #217: Consider adding From<uN> for HeaderValue
It's incredibly common to need to set integer values in headers, but it's currently not super intuitive how to do that. You essentially need to format your integer into a string, and then use `HeaderValue::from_str`. This requires `HeaderValue` to parse the string for invalid bytes, which is most definitely valid since ASCII decimal characters are valid. A naive approach from a user will also mean copying the bytes a couple times from a `String` into headervalue's internal `Bytes`.

We could consider adding `From<usize>`, `From<u64>`, etc for `HeaderValue`, since we know already that it would be a valid header value. And it would also be a big improvement when using the builders:

```rust
Response::builder()
    .header("content-length", buf.len())
    .body(buf)
```

seanmonstar:
This seems super useful, but I could be convinced that it shouldn't be in-crate.

Closes #217

### DIFF
--- a/src/header/value.rs
+++ b/src/header/value.rs
@@ -1,6 +1,6 @@
 use bytes::{Bytes, BytesMut};
 
-use std::{cmp, fmt, str};
+use std::{cmp, fmt, mem, str};
 use std::error::Error;
 use std::str::FromStr;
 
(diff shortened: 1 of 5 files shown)


seanmonstar:
I just thought this, on 32bit systems, serializing a `u64` will require `BytesMut` to allocate unconditionally (since max decimal length if 20, and inline bytes on 32bit is 15). This could be updated to include a `#[cfg(target_pointer_width = "32")]` version of u64/i64 that checks if the value is bigger than a constant of `10.pow(15) - 1`, and only in the rare cases would it then need to allocate.

src/header/value.rs
```diff
@@ -359,6 +360,61 @@ impl From<HeaderName> for HeaderValue {
...
+
+from_integers! {
+    // integer type => maximum decimal length
+
+    // u8 purposely left off... HeaderValue::from(b'3') could be confusing
+    u16 => 5,
+    i16 => 6,
+    u32 => 10,
+    i32 => 11,
+    u64 => 20,
```
carllerche: As you mentioned, this would allocate on 32bit systems. In these cases, you could probably compute the real length using the number value.

Issue #206, opened 2018-06-02
There should be an uri::Builder


SergejJurecko:
Right now the burden of correct percent encoding is entirely on the user. This is completely unnecessary. Only accepting string URI also forces users to concatenate strings to construct an URI, which is bad and inefficient practice that leads to bugs.

## Examples

Entirely segmented 
`
Uri::builder()
        .https()    // In mio_httpc I have both https() and set_https(true).
        .host("www.example.com")
        .auth("user name", "my password")
        .port(12345)
        .path_segm("a/")
        .path_segm("b")
        .query("spaced key", "123")
        .query("key", "<>")
`

Only partly segmented because schemes, hosts and optional ports are not a percent encoding problem and often stored as a fixed string on a client.

`
Uri::builder()
        .base("https://www.example.com:8080")
        .path_segm("a/")
        .path_segm("b")
        .query("spaced key", "123")
        .query("key", "<>")
`

Query and path segments could also be slices.

`
Uri::builder()
        .base("https://www.example.com:8080")
        .path(&["a", "b", "c d"])
        .query_pairs(&[("spaced key", "123"),("another","key")])
`

seanmonstar:
Thanks for writing this up! I agree with premise, we should be able to provide a builder for easier creation.

Since different parts of a URI have different allowed characters, the builder would need to return a `Result`.

SergejJurecko:
Invalid characters are percent encoded so I don't see a way for query/path functions to fail.  `.path_segm("a/")` should percent encode the slash. Base should return Result<Self> because you can't stop a user from writing nonsense.

seanmonstar:
It doesn't have all that is mentioned in here, but I've at least *started* a builder concept in https://github.com/hyperium/http/pull/219.

seanmonstar:
#219 has been merged, which provides methods for:

- `scheme`
- `authority`
- `path_and_query`

There is room to add methods for smaller parts, but they should be discussed first.

yageek:
To add more granularities to the Builder, some additionnal struct could be addeded. The PR #268 proposes a `QueryItem` struct to stores the information about the query component path. This is directly inspired from Swift [URLQueryItem](https://developer.apple.com/documentation/foundation/urlqueryitem). It has one string field for the key and one optional string for the value.

If the builder would allow to build path and query separately, we could offer a better way, according to me, to abstract the different endpoints of an API.

[...]

PR #219, opened 2018-06-22
Introduce a Uri Builder

seanmonstar:
This introduces a `Builder` type for manipulating and building up `Uri`s. Instead of having mutators on `Uri` directly, which would need to return errors, instead this suggests that a user either use a `Builder` from the start, or to convert a `Uri` into via `uri.into_builder()`.

An example looks like:

```rust
let uri = Uri::builder()
    .scheme("https")
    .authority("hyper.rs")
    .path_and_query("/")
    .build()
    .unwrap();
```

It currently only supports setting the 3 distinct parts, not any sub parts (like host or path), but I expect those could be added. This is however generic over `HttpTryFrom`, so you can easily build pieces with `&str`, or `Bytes`, etc.

It's only a start, but related to #206.

### DIFF
--- a/src/convert.rs
+++ b/src/convert.rs
@@ -3,7 +3,7 @@ use header::{HeaderName, HeaderValue};
 use method::Method;
 use sealed::Sealed;
 use status::StatusCode;
-use uri::Uri;
+use uri::{Scheme, Authority, PathAndQuery, Uri};
 
 /// Private trait for the `http` crate to have generic methods with fallible
 /// conversions.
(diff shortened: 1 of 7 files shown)


bluetech:
`Uri` as described in the docs has 4 distinct forms:

```
request-target = origin-form
               / absolute-form
               / authority-form
               / asterisk-form
```

Maybe it makes sense to have 4 distinct builders, one for each type? (Though the asterisk form doesn't really need one).

src/uri/mod.rs
```diff
@@ -289,6 +296,11 @@ impl Uri {
         parse_full(s)
     }
 
+    /// dox
+    pub fn into_builder(self) -> Builder {
```
carllerche: Maybe leave this fn off for now?

src/convert.rs
```diff
@@ -22,6 +22,22 @@ pub trait HttpTryFrom<T>: Sized + Sealed {
     fn try_from(t: T) -> Result<Self, Self::Error>;
 }
 
+pub(crate) trait HttpTryInto<T>: Sized {
```
carllerche: Is this added for internal convenience or does it impact the public api somehow.

seanmonstar: Yea, it was internal convenience, allowing `arg.http_try_into()` instead of `<T as HttpTryFrom<U>>::try_from(arg)`. Since the trait isn't public, and no one can import it, it exposes no public contract.

seanmonstar:
@carllerche took a while, but I finally cleaned this up.

PR #220, opened 2018-06-23
Fix `Uri::from_str("S")` from parsing to a `S/` Uri

### LINKED ISSUE #197: Panic in Uri::from_str(&Uri::from_str(x).to_string()) found by fuzzer
In <https://github.com/rust-fuzz/targets/pull/113>, I added some simple fuzzers for parts of this crate. To test `Uri::from_str`, I added this code:

```rust
pub fn fuzz_http_uri(data: &[u8]) {
    use std::str;
    use std::str::FromStr;

    if let Ok(text) = str::from_utf8(data) {
        if let Ok(uri) = http::uri::Uri::from_str(text) {
            http::uri::Uri::from_str(&uri.to_string()).unwrap();
        }
    }
}
```

- Is this a correct assertion to make?
- This panics with `b"S"` as an input

seanmonstar:
Closes #197

### DIFF
--- a/src/uri/mod.rs
+++ b/src/uri/mod.rs
@@ -269,8 +269,7 @@ impl Uri {
                         return Ok(Uri {
                             scheme: Scheme::empty(),
                             authority: authority,
-                            // TODO: Should this be empty instead?
-                            path_and_query: PathAndQuery::slash(),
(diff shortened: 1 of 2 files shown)


seanmonstar:
Thanks nightly...

Issue #87, opened 2017-08-01
export http::HeaderValue?


olix0r:
this crate exports `http::HeaderMap` and `http::header::HeaderValue`.  The typical signature of HeaderMap is `HeaderMap<HeaderValue>`, so it seems natural for both of these to exported together from the root of the module hierarchy.

carllerche:
I'm going to remove the 0.1 milestone as this change would not be breaking.

PR #222, opened 2018-06-24
Export http::header::HeaderValue as http::HeaderValue

joshleeb:
Export `http::header::HeaderValue` as `http::HeaderValue` and update benchmarks and doc tests.

Closes: #87


seanmonstar:
Thanks for the PR!

I wonder if this is really needed. In all the places I've used `http::HeaderMap`, I've never needed to quality that it is a `HeaderMap<HeaderValue>`. And instead, it seems useful to keep this import as `http::header::HeaderValue`... But I could be wrong...

carllerche:
I'm pro re-exporting it.

I've definitely wanted this a few times, but not to qualify `HeaderMap`.. mostly that `HeaderValue` is proving to be a pretty common type to use and having it at the root seems useful. Also, because the type name includes "Header", it seems safe enough to include in the root (lack of ambiguity).

I think that I find the stutter most annoying (`header::Header...`)

Not at all a comprehensive list (even in the projects listed), but more of a sample:

### Tower Web

* https://github.com/carllerche/tower-web/blob/master/src/response/default_serializer.rs#L6
* https://github.com/carllerche/tower-web/blob/master/tower-web-macros/src/derive/attr.rs#L1-L2

### h2

* https://github.com/carllerche/h2/blob/master/src/hpack/header.rs#L4-L5
* https://github.com/carllerche/h2/blob/master/src/frame/headers.rs#L6
* https://github.com/carllerche/tower-web/blob/master/src/response/content_type.rs#L1

seanmonstar:
@carllerche do you still feel that this should be merged? If so, go ahead, otherwise we should close and not leave dangling (I still personally feel its fine to be imported via `http::header::HeaderValue`, but don't care to block).

carllerche:
I would merge

carllerche:
I fixed the merge conflict. Once CI passes, I will merge it.

PR #231, opened 2018-07-28
Allow user to use Parts

DoumanAsh:
I would like to be able to write own builder effectively as `http::Builder` is pretty limited.

For this reason I believe user should be able to create own instance of parts which then could be used to instantiate `Request` using it and body.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -158,6 +158,7 @@ pub struct Request<T> {
     body: T,
 }
 
+#[derive(Default)]
 /// Component parts of an HTTP `Request`
 ///
 /// The HTTP request head consists of a method, uri, version, and a set of
@@ -177,8 +178,6 @@ pub struct Parts {
 
     /// The request's extensions
     pub extensions: Extensions,
-
-    _priv: (),
 }
 
 /// An HTTP request builder
@@ -423,7 +422,7 @@ impl<T> Request<T> {
     #[inline]
     pub fn new(body: T) -> Request<T> {
         Request {
-            head: Parts::new(),
+            head: Parts::default(),
             body: body,
         }
     }
@@ -697,20 +696,6 @@ impl<T: fmt::Debug> fmt::Debug for Request<T> {
     }
 }
 
-impl Parts {
-    /// Creates a new default instance of `Parts`
-    fn new() -> Parts {
-        Parts{
-            method: Method::default(),
-            uri: Uri::default(),
-            version: Version::default(),
-            headers: HeaderMap::default(),
-            extensions: Extensions::default(),
-            _priv: (),
-        }
-    }
-}
-
 impl fmt::Debug for Parts {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         f.debug_struct("Parts")
@@ -939,7 +924,7 @@ impl Default for Builder {
     #[inline]
     fn default() -> Builder {
         Builder {
-            head: Some(Parts::new()),
+            head: Some(Parts::default()),
             err: None,
         }
     }


seanmonstar:
The API was actually designed specifically to **prevent** this. Removing the `_priv` field would make it impossible to add any new fields in the future.

You mentioned in Gitter that "the current Builder really sucks", would you be interested in the current builder being improved? What issues to do you have with it?

DoumanAsh:
Specifically the current set of methods very much limits its usability.
For example I can only append headers, I cannot check whether header already present and replace it if I need.

The simplest suggestion would for `Builder` to provide inner fields of `Parts` so that user could wrap builder in its own.
I currently use my branch to basically write all functionality I need and provide myself easy way to construct request from parts and body

There is currently no way to extend `http::Builder` without rewriting it from a scratch.
Using `Parts` seemed to me like a perfect candidate for user to write own builder.

My current interface looks like this https://gist.github.com/DoumanAsh/3b7e8d1f878653402ba75cd31cf409a2

But it is pretty much work in progress until I'll hit places where I'd need to extend it more.

DoumanAsh:
Yeah, at basic level having mutable references to all components should be more than enough to extend builder

seanmonstar:
I forgot, a potential blocker is that the builder could already be in an error state, which prevents returning a mutable reference to the internal part...

DoumanAsh:
Ah yes... I forgot about this unpleasant part.

In this case I feel there should be a way for user to create `Parts` that can be used to create `Request`.
How about there would be `Default` impl at least for `Parts`?

Of course without getting rid of `priv` field?
As its fields are already public, I believe there should be no issue?

DoumanAsh:
It would require user to have some cheap default implementation for Request's body type.
Not to mention, it defeats purpose of Builder itself as it is pretty much limited comparing to `Request` itself.
You might as well deprecate it and obsolete, if users should use Request directly.

Considering there is [Request::from_parts](https://docs.rs/http/0.1.8/http/request/struct.Request.html#method.from_parts) that could let me just use `Parts` as it is, It seems kinda excessive to automatically have `Default` body when the user could use common `Parts` to setup request and then build it by combining with provided `Body`

`Parts` are already public and their components are exposed
I do not see a reason not let user to create it.

Even if the http crate made in mind for hyper only, it shouldn't prevent people from using all of it components.
Though if you feel like it is bad idea for whatever reason, feel free to close PR.

carllerche:
None of your latest comment makes any sense to me.

You asked for a way to implement a custom builder. I suggested:

```rust
struct MyBuilder {
   request: Request<()>,
}

impl MyBuilder {
   // whatever you want here

   pub fn body<T>(self, body: T) -> Request<T> {
      self.request.map(|_| body)
   }
}
```

> it defeats purpose of Builder itself as it is pretty much limited comparing to Request itself.

The purpose of the builder is convenience, not flexibility. As you pointed out, using the `Request` directly has more power, but it requires much more verbosity.

> You might as well deprecate it and obsolete, if users should use Request directly.

We will take this under advisement.

Anyway, this PR does not fit in with the goals of the `http` crate, so I am going to close it. That said, what you are trying to achieve, i.e. "I would like to be able to write own builder" is already possible and does not require changes to the library.

DoumanAsh:
@carllerche Having public `Parts` and no way to create it aside from exacting it out of existing Request, makes no sense to me too. JFYI

>That said, what you are trying to achieve, i.e. "I would like to be able to write own builder" is already possible and does not require changes to the library.

Yeah, of course it is possible, and it may be even easy if you use `hyper` as its body have cheap default variant. Or if user uses own `Body` that has cheap default impl

carllerche:
> Yeah, of course it is possible, and it may be even easy if you use hyper as its body have cheap default variant. Or if user uses own Body that has cheap default impl

Did you read the example I posted above? Based on this comment, I do not think you understood. The cheap default implementation is completely unrelated to how you can implement your own builder.

DoumanAsh:
@seanmonstar I actually settled down on using `Request::into_parts()` to just get parts as it is supposed to be done 😄

Issue #202, opened 2018-05-11
Double % percent encoding crashes with ErrorKind::InvalidUriChar


ernestas-poskus:
```rust
extern crate http;

use http::Uri;

fn main() {
    let val = "/your.js?bn=%%ADFTAGID%%;";

    println!("ok {:?}", val.parse::<Uri>().unwrap());
}
```

Uri fails to parse double `%%` percent while Url crate (https://github.com/servo/rust-url) and Hyper `0.11.26` internal implementation of Uri succeds https://github.com/hyperium/hyper/blob/v0.11.26/src/uri.rs
https://github.com/hyperium/http/blob/master/src/uri/path.rs#L54

Looks like the problem is below:
```rust
                        let perc_encoded = i + 3 <= src.len() && // enough capacity
                            HEX_DIGIT[src[i + 1] as usize] != 0
                            && HEX_DIGIT[src[i + 2] as usize] != 0;
```

I did some debugging

```rust
            println!("GOT i: {} URI_CHARS [{}, {}]", i, b, URI_CHARS[b as usize]);

            println!("SRC +1 {} +2 {}", src[i + 1], src[i + 2],);
```

```
GOT i: 0 URI_CHARS [47, 47]
SRC +1 121 +2 111
GOT i: 1 URI_CHARS [121, 121]
SRC +1 111 +2 117
GOT i: 2 URI_CHARS [111, 111]
SRC +1 117 +2 114
GOT i: 3 URI_CHARS [117, 117]
SRC +1 114 +2 46
GOT i: 4 URI_CHARS [114, 114]
SRC +1 46 +2 106
GOT i: 5 URI_CHARS [46, 46]
SRC +1 106 +2 115
GOT i: 6 URI_CHARS [106, 106]
SRC +1 115 +2 63
GOT i: 7 URI_CHARS [115, 115]
SRC +1 63 +2 98
GOT i: 8 URI_CHARS [63, 63]
SRC +1 98 +2 110
GOT i: 9 URI_CHARS [98, 98]
SRC +1 110 +2 61
GOT i: 10 URI_CHARS [110, 110]
SRC +1 61 +2 37
GOT i: 11 URI_CHARS [61, 61]
SRC +1 37 +2 37
GOT i: 12 URI_CHARS [37, 0]
SRC +1 37 +2 65
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidUri(InvalidUriChar)', libcore/result.rs:945:5
```

[...]

PR #233, opened 2018-07-31
Fix path double percent parsing

ernestas-poskus:
Issue: https://github.com/hyperium/http/issues/202

According to RFC https://tools.ietf.org/html/rfc3986#section-2.1

Percent character "%" must be followed by the two hexadecimal digits when is encoded - therefore if percent character is immediately followed by percent character conditional does not hold.

### DIFF
--- a/src/uri/path.rs
+++ b/src/uri/path.rs
@@ -49,6 +49,11 @@ impl PathAndQuery {
             match URI_CHARS[b as usize] {
                 0 => {
                     if b == b'%' {
+                        // Check if next character is not %
+                        if i + 2 <= src.len() && b'%' == src[i + 1] {
+                            break;
+                        }
+
                         // Check that there are enough chars for a percent
                         // encoded char
                         let perc_encoded =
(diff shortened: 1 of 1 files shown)


src/uri/path.rs
```diff
@@ -49,6 +49,11 @@ impl PathAndQuery {
             match URI_CHARS[b as usize] {
                 0 => {
                     if b == b'%' {
+                        // Check if next character is not %
+                        if b'%' == src[i + 1] {
```
seanmonstar: I believe this will panic if `%` were the last character, indexing out of bounds.

ernestas-poskus: classic, thanks for noticing! fixed and added additional test

src/uri/path.rs
```diff
@@ -49,6 +49,11 @@ impl PathAndQuery {
             match URI_CHARS[b as usize] {
                 0 => {
                     if b == b'%' {
+                        // Check if next character is not %
+                        if i + 1 > src.len() && b'%' == src[i + 1] {
```
peddermaster2: Shouldn't it be: 
```rust
if i + 1 >= src.len() || b'%' == src[i + 1] {
```

seanmonstar:
Er, woops, looks like test failures on CI are legitimate.

src/uri/path.rs
ernestas-poskus: apparently should have been +2, why `||` OR ?

src/uri/path.rs
```diff
@@ -504,4 +509,32 @@ mod tests {
...
+    #[test]
+    fn double_percent_path() {
+        let double_percent_path = "/your.js?bn=%%val";
+
+        assert!(double_percent_path.parse::<PathAndQuery>().is_ok());
+
+        let path: PathAndQuery = double_percent_path.parse().unwrap();
+        assert_eq!(path, double_percent_path);
+
+        let double_percent_path = "/path%%";
```
ernestas-poskus: I have added few additional test cases now if path ends with double percent no error is returned on the other hand if ends with 1 percent `ErrorKind::InvalidUriChar` error is returned

src/uri/path.rs
```diff
@@ -504,4 +509,32 @@ mod tests {
...
+
+        assert!(double_percent_path.parse::<PathAndQuery>().is_ok());
+    }
+
+    #[test]
+    fn path_ends_with_question_mark() {
+        let path = "/path?%";
```
ernestas-poskus: I path ends with % it will return error ErrorKind::InvalidUriChar

seanmonstar:
Thanks!

PR #234, opened 2018-08-06
Add HttpTryFrom impls for &Self

carllerche:


src/header/name.rs
```diff
@@ -1742,6 +1742,15 @@ impl From<HeaderName> for Bytes {
     }
 }
 
+impl<'a> HttpTryFrom<&'a HeaderName> for HeaderName {
+    type Error = InvalidHeaderName;
```
seanmonstar: Should the error types of these be `Never` instead?

carllerche: Probably could. I forgot we added a `Never` type.

seanmonstar: Unfortunately, it might be less symmetrical, since I think `HttyTryfrom<HeaderName> for HeaderName` and friends have respective error types, instead of `Never`... I'm not sure which is better. Fixing *new* impls as their added, or being consistent and changing *all* impls in 0.2.

carllerche:
I switched to `Never`. Either way, it seems pretty minor.

PR #239, opened 2018-08-15
fix Send / Sync (typos)

sparecycles:
I think I found a couple copy-paste errors.

### DIFF
--- a/src/header/name.rs
+++ b/src/header/name.rs
@@ -2024,7 +2024,7 @@ mod tests {
 
     #[test]
     fn test_bounds() {
-        fn check_bounds<T: Sync + Send>() {}
+        fn check_bounds<T: Send + Sync>() {}
         check_bounds::<HeaderName>();
     }
 
--- a/src/header/map.rs
+++ b/src/header/map.rs
@@ -1939,7 +1939,7 @@ impl<'a, T> Iterator for Iter<'a, T> {
 }
 
 unsafe impl<'a, T: Sync> Sync for Iter<'a, T> {}
-unsafe impl<'a, T: Send> Send for Iter<'a, T> {}
+unsafe impl<'a, T: Sync> Send for Iter<'a, T> {}
 
 // ===== impl IterMut =====
 
--- a/src/header/map.rs
+++ b/src/header/map.rs
@@ -1939,7 +1939,7 @@ impl<'a, T> Iterator for Iter<'a, T> {
 }
 
 unsafe impl<'a, T: Sync> Sync for Iter<'a, T> {}
-unsafe impl<'a, T: Sync> Send for Iter<'a, T> {}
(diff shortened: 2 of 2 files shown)


src/header/name.rs
```diff
@@ -2024,7 +2024,7 @@ mod tests {
 
     #[test]
     fn test_bounds() {
-        fn check_bounds<T: Sync + Send>() {}
+        fn check_bounds<T: Send + Sync>() {}
```
sparecycles: every other usage is in `Send + Sync` order.

carllerche:
How does HashMap in std implement Send and Sync bounds?

carllerche:
I actually don’t think that Iter can be Send because it holds a reference to the map. If I recall those bounds were intentional.

sparecycles:
Hi @carllerche.  I'm not sure if Iter should be Send or not.

But if that's the case, wouldn't Drain, IterMut, and ValueIterMut have the same bounds?

https://github.com/hyperium/http/blob/8d45bd56007ef341686a54cf092446fe8f988123/src/header/map.rs#L2002-L2003

https://github.com/hyperium/http/blob/8d45bd56007ef341686a54cf092446fe8f988123/src/header/map.rs#L2102-L2103

https://github.com/hyperium/http/blob/8d45bd56007ef341686a54cf092446fe8f988123/src/header/map.rs#L2531-L2532

---

...

https://github.com/hyperium/http/blob/8d45bd56007ef341686a54cf092446fe8f988123/src/header/map.rs#L1941-L1942

sparecycles:
@carllerche you're right! This pattern is also in the rust core libraries.  Still feels like a typo (and if not, that it needs a nice fat comment).

https://github.com/rust-lang/rust/blob/917945d662c42053383fe3e71cb0f313d585e459/src/liballoc/collections/linked_list.rs#L1202-L1218

https://github.com/rust-lang/rust/blob/917945d662c42053383fe3e71cb0f313d585e459/src/libcore/slice/mod.rs#L2650-L2653

https://github.com/rust-lang/rust/blob/master/src/libstd/collections/hash/table.rs#L916-L917

sparecycles:
@carllerche You're right, the pattern of [sync-implies-ref-is-send] is deep in the rust type system.

[sync-implies-ref-is-send]: https://github.com/rust-lang/rust/commit/35ca50bd5676db31a8074a216d1aadad7d434de8

carllerche:
Would you mind adding a comment including your findings for the next person?

sparecycles:
Well, I think I understand why `Iter<T: Sync>` is `Sync + Send`, but I don't understand why `Iter<T: Send>` is not `Send`, nor why `IterMut`/`Drain` etc... don't follow the same pattern.

If a good explanation appears in [rust#53625](https://github.com/rust-lang/rust/issues/53625), I'll can open this with a nice comment.

PR #240, opened 2018-08-20
Add `Uri::from_static`

tesaguri:
This is useful for creating a `Uri` from a string literal.

### DIFF
--- a/src/uri/mod.rs
+++ b/src/uri/mod.rs
@@ -288,6 +288,32 @@ impl Uri {
         parse_full(s)
     }
 
+    /// Convert a `Uri` from a static string.
+    ///
+    /// This function will not perform any copying, however the string is
+    /// checked to ensure that it is valid.
+    ///
+    /// # Panics
+    ///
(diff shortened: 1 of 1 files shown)


src/uri/mod.rs
```diff
@@ -288,6 +288,29 @@ impl Uri {
...
+    /// ```
+    /// # use http::uri::Uri;
+    /// let uri = Uri::from_static("http://example.com/foo");
+    ///
+    /// assert_eq!(uri.host().unwrap(), "example.com");
+    /// assert_eq!(uri.path(), "/foo");
+    /// ```
+    pub fn from_static(src: &'static str) -> Self {
+        let s = Bytes::from_static(src.as_bytes());
+        Uri::from_shared(s).unwrap()
```
seanmonstar: It may be useful to actually match on this and `panic!` with a clear message, so people see if came from `Uri::from_static`, and not `Result::unwrap` in libstd.

carllerche:
Lgtm

PR #246, opened 2018-09-10
Update quickcheck to 0.7

ignatenkobrain:


ignatenkobrain:
Didn't notice that it breaks things. We need update to rand 0.5 as well...

PR #252, opened 2018-09-14
Add and use a Port struct for Uri::port()

### LINKED ISSUE #173: Consider a Port type for Uri::port()
Users may wish to access the port as a string, and not just an integer. We have the string already, so we could provide a newtype around it, `Port`, and a user can then get the bytes themselves, or just ask for the integer.

```rust
fn port(&self) -> Option<Port>;
```

```rust
struct Port<'a> {
    bytes: &'a str,
}

impl<'a> Port<'a> {
    fn as_str(&self) -> &str;

    fn as_u16(&self) -> u16;
}
```

This would be for both `Uri` and `Authority`. Probably introduced as `port_part()`, deprecating `port()` to be replaced in 0.2.

joshleeb:
Add the `uri::Port` struct an use it with `Authority::port_path`, and `Uri::port_path` methods, which deprecate `Authority::port` and `Uri::port` methods.

Closes #173 - Consider a Port type for `Uri::port()`

### DIFF
--- a/src/uri/authority.rs
+++ b/src/uri/authority.rs
@@ -188,11 +188,10 @@ impl Authority {
 
     /// Get the port part of this `Authority`.
     ///
-
     /// The port subcomponent of authority is designated by an optional port
-    /// number as bytes following the host and delimited from it by a single
-    /// colon (":") character. It can be turned into a decimal port number with
-    /// the `as_u16` method.
+    /// number following the host and delimited from it by a single colon (":")
+    /// character. It can be turned into a decimal port number with the `as_u16`
+    /// method or as a `str` with the `as_str` method.
     ///
     /// ```notrust
     /// abc://username:password@example.com:123/path/data?key=value&key2=value2#fragid1
--- a/src/uri/authority.rs
+++ b/src/uri/authority.rs
@@ -8,7 +8,7 @@ use std::str::FromStr;
 use bytes::Bytes;
 
 use byte_str::ByteStr;
-use super::{ErrorKind, InvalidUri, InvalidUriBytes, URI_CHARS};
+use super::{ErrorKind, InvalidUri, InvalidUriBytes, URI_CHARS, Port};
 
 /// Represents the authority component of a URI.
 #[derive(Clone)]
@@ -180,12 +180,19 @@ impl Authority {
         host(self.as_str())
     }
 
-    /// Get the port of this `Authority`.
+    #[deprecated(since="0.2.0", note="please use `port_part` instead")]
+    #[doc(hidden)]
+    pub fn port(&self) -> Option<u16> {
+        self.port_part().and_then(|p| Some(p.as_u16()))
+    }
+
+    /// Get the port part of this `Authority`.
     ///
+
     /// The port subcomponent of authority is designated by an optional port
-    /// number in decimal following the host and delimited from it by a single
-    /// colon (":") character. A value is only returned if one is specified in
-    /// the URI string, i.e., default port values are **not** returned.
+    /// number as bytes following the host and delimited from it by a single
+    /// colon (":") character. It can be turned into a decimal port number with
+    /// the `as_u16` method.
     ///
     /// ```notrust
     /// abc://username:password@example.com:123/path/data?key=value&key2=value2#fragid1
@@ -202,7 +209,8 @@ impl Authority {
     /// # use http::uri::Authority;
     /// let authority: Authority = "example.org:80".parse().unwrap();
     ///
-    /// assert_eq!(authority.port(), Some(80));
+    /// let port = authority.port_part().unwrap();
+    /// assert_eq!(port.as_u16(), 80);
     /// ```
     ///
     /// Authority without port
@@ -211,13 +219,13 @@ impl Authority {
     /// # use http::uri::Authority;
     /// let authority: Authority = "example.org".parse().unwrap();
     ///
-    /// assert!(authority.port().is_none());
+    /// assert!(authority.port_part().is_none());
     /// ```
-    pub fn port(&self) -> Option<u16> {
-        let s = self.as_str();
-        s.rfind(":").and_then(|i| {
-            u16::from_str(&s[i+1..]).ok()
-        })
+    pub fn port_part(&self) -> Option<Port> {
+        let bytes = self.as_str();
+        bytes
+            .rfind(":")
+            .and_then(|i| Port::from_str(&bytes[i + 1..]).ok())
     }
 
     /// Return a str representation of the authority
--- a/src/uri/mod.rs
+++ b/src/uri/mod.rs
@@ -554,9 +554,9 @@ impl Uri {
     /// Get the port part of this `Uri`.
     ///
     /// The port subcomponent of authority is designated by an optional port
-    /// number as bytes following the host and delimited from it by a single
-    /// colon (":") character. It can be turned into a decimal port number with
-    /// the `as_u16` method.
+    /// number following the host and delimited from it by a single colon (":")
+    /// character. It can be turned into a decimal port number with the `as_u16`
+    /// method or as a `str` with the `as_str` method.
     ///
     /// ```notrust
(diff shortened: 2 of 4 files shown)


src/uri/authority.rs
```diff
@@ -180,12 +180,19 @@ impl Authority {
         host(self.as_str())
     }
 
-    /// Get the port of this `Authority`.
+    /// Get the port of this `Authority`. Deprecated: see `port_path` instead.
+    #[deprecated(since="0.2.0", note="please use `port_path` instead")]
+    pub fn port(&self) -> Option<u16> {
```
carllerche: Could this be doc hidden as well?

src/uri/authority.rs
```diff
@@ -196,28 +203,28 @@ impl Authority {
     ///
     /// # Examples
     ///
-    /// Authority with port
+    /// Authority with port path
```
carllerche: I would keep the docs the same. No need to add “part”.

src/uri/authority.rs
```diff
@@ -196,28 +203,28 @@ impl Authority {
...
     ///
-    /// assert!(authority.port().is_none());
+    /// assert!(authority.port_path().is_none());
     /// ```
-    pub fn port(&self) -> Option<u16> {
-        let s = self.as_str();
-        s.rfind(":").and_then(|i| {
-            u16::from_str(&s[i+1..]).ok()
-        })
+    pub fn port_path(&self) -> Option<Port> {
```
carllerche: We have been using a _part suffix for this situation. We probably want to stick with that pattern.

src/uri/mod.rs
```diff
@@ -543,12 +545,18 @@ impl Uri {
         self.authority_part().map(|a| a.host())
     }
 
-    /// Get the port of this `Uri`.
+    /// Get the port of this `Uri`. Deprecated: see `port_path` instead.
+    #[deprecated(since="0.2.0", note="please use `port_path` instead")]
+    pub fn port(&self) -> Option<u16> {
```
carllerche: Same re doc hidden.

src/uri/port.rs
```diff
@@ -0,0 +1,67 @@
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use http::uri::Port;
+    ///
+    /// assert!(Port::from_str("80").is_ok());
+    /// assert!(Port::from_str("abcd").is_err());
+    /// ```
+    pub fn from_str(bytes: &'a str) -> Result<Self, num::ParseIntError> {
```
carllerche: I would not expose this fn yet in favor of not letting the type be publicly constructed yet.

joshleeb:
@carllerche updated :)

src/uri/authority.rs
```diff
@@ -180,12 +180,19 @@ impl Authority {
         host(self.as_str())
     }
 
-    /// Get the port of this `Authority`.
+    #[deprecated(since="0.2.0", note="please use `port_path` instead")]
```
carllerche: Stray `_path` reference.

src/uri/authority.rs
```diff
@@ -180,12 +180,19 @@ impl Authority {
...
+    }
+
+    /// Get the port part of this `Authority`.
     ///
-    /// The port subcomponent of authority is designated by an optional port
-    /// number in decimal following the host and delimited from it by a single
-    /// colon (":") character. A value is only returned if one is specified in
-    /// the URI string, i.e., default port values are **not** returned.
+
+    /// The port path subcomponent of authority is designated by an optional port
```
carllerche: Stray "path" reference.

src/uri/authority.rs
```diff
@@ -199,25 +206,25 @@ impl Authority {
     /// Authority with port
     ///
     /// ```
-    /// # use http::uri::Authority;
+    /// # use http::uri::{Authority, Port};
     /// let authority: Authority = "example.org:80".parse().unwrap();
     ///
-    /// assert_eq!(authority.port(), Some(80));
+    /// assert_eq!(authority.port_path(), Port::from_str("80").ok());
```
carllerche: _path reference

src/uri/authority.rs
```diff
@@ -199,25 +206,25 @@ impl Authority {
...
     ///
     /// Authority without port
     ///
     /// ```
-    /// # use http::uri::Authority;
+    /// # use http::uri::{Authority, Port};
     /// let authority: Authority = "example.org".parse().unwrap();
     ///
-    /// assert!(authority.port().is_none());
+    /// assert!(authority.port_path().is_none());
```
carllerche: _path reference.

joshleeb:
Ahh damn, sorry.

@carllerche should be fixed up properly now

src/uri/port.rs
```diff
@@ -0,0 +1,64 @@
...
+    bytes: &'a str,
+    port: u16,
+}
+
+impl<'a> Port<'a> {
+    /// Converts a `str` to a port number.
+    ///
+    /// The supplied `str` must be a valid u16.
+    pub(crate) fn from_str(bytes: &'a str) -> Result<Self, num::ParseIntError> {
+        u16::from_str(bytes).and_then(|port| Ok(Port { port, bytes }))
```
bluetech: The BNF of a URI port is `*DIGIT`, see https://tools.ietf.org/html/rfc3986#section-3.2.3. However `from_str` allows an optional `+` sign, see https://doc.rust-lang.org/nightly/std/primitive.u16.html#method.from_str_radix.

carllerche: In theory, this should already have been validated when constructing the `Uri`, so checking again here is not necessary.

src/uri/authority.rs
```diff
@@ -180,12 +180,19 @@ impl Authority {
...
+    }
+
+    /// Get the port part of this `Authority`.
     ///
+
     /// The port subcomponent of authority is designated by an optional port
-    /// number in decimal following the host and delimited from it by a single
-    /// colon (":") character. A value is only returned if one is specified in
-    /// the URI string, i.e., default port values are **not** returned.
+    /// number as bytes following the host and delimited from it by a single
```
carllerche: nit: I would remove `as bytes` here as that is an implementation detail.

src/uri/authority.rs
```diff
@@ -180,12 +180,19 @@ impl Authority {
...
+    /// colon (":") character. It can be turned into a decimal port number with
+    /// the `as_u16` method.
```
carllerche: Maybe also add "or as a `str` with the `as_str` method.

src/uri/mod.rs
```diff
@@ -543,12 +545,18 @@ impl Uri {
...
+        self.port_part().and_then(|p| Some(p.as_u16()))
+    }
+
+    /// Get the port part of this `Uri`.
     ///
     /// The port subcomponent of authority is designated by an optional port
-    /// number in decimal following the host and delimited from it by a single
-    /// colon (":") character. A value is only returned if one is specified in
-    /// the URI string, i.e., default port values are **not** returned.
+    /// number as bytes following the host and delimited from it by a single
```
carllerche: Same doc nits as above.

src/uri/port.rs
```diff
@@ -0,0 +1,64 @@
...
+    ///
+    /// let port = authority.port_part().unwrap();
+    /// assert_eq!(port.as_str(), "80");
+    /// ```
+    pub fn as_str(&self) -> &'a str {
+        self.bytes
+    }
+}
+
+impl<'a> PartialEq for Port<'a> {
```
carllerche: Not required for this PR, but it would be nice to add `PartialEq` implementations for numeric types (`u16`, `u32`, ...).

seanmonstar:
Looks good to me when nits are fixed. Thanks!

src/uri/port.rs
joshleeb: I'll put up a subsequent PR adding a generic implementation for `PartialEq` 😄

PR #255, opened 2018-09-20
Add trait impls for uri::Port

joshleeb:
Implement some useful trait impls for `uri::Port`. This follows on from #252 and specifically [this comment](https://github.com/hyperium/http/pull/252#discussion_r218481297).

### DIFF
--- a/src/uri/port.rs
+++ b/src/uri/port.rs
@@ -20,7 +20,7 @@ impl<'a> Port<'a> {
     ///
     /// # Examples
     ///
-    /// Port as `u16`
+    /// Port as `u16`.
     ///
     /// ```
     /// # use http::uri::Authority;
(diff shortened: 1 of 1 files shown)


src/uri/port.rs
```diff
@@ -51,14 +51,32 @@ impl<'a> Port<'a> {
...
 impl<'a> PartialEq for Port<'a> {
     fn eq(&self, other: &Self) -> bool {
         self.port == other.port
     }
 }
 
-impl<'a> fmt::Display for Port<'a> {
-    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
-        write!(fmt, "{}", self.bytes)
+impl<'a, T: From<u16> + PartialEq> PartialEq<T> for Port<'a> {
```
carllerche: I would rather avoid a blanket impl for this partialeq. I believe you did this to avoid having to add impls for each u* type. Perhaps a macro would work instead?

The problem is that adding this blanket impl would prevent us from adding a different blanket impl. Perhaps we would like to `impl PartialEq<T>` for `T: AsRef<Port>` or something.

src/uri/port.rs
```diff
@@ -51,14 +51,32 @@ impl<'a> Port<'a> {
...
+    }
+}
+
+impl<'a> Into<u16> for Port<'a> {
+    fn into(self) -> u16 {
+        self.as_u16()
+    }
+}
+
+impl<'a> Into<&'a str> for Port<'a> {
```
carllerche: It also looks like we could impl `AsRef<str>` for `Port`.

joshleeb:
@carllerche updated 😄

src/uri/port.rs
```diff
@@ -51,14 +51,99 @@ impl<'a> Port<'a> {
...
+            }
+        }
+    }
+}
+
+port_partialeq!(u16);
+port_partialeq!(u32);
+port_partialeq!(u64);
+
+port_partialeq!(i16);
```
carllerche: At a glance, an unchecked cast from `u16` -> `i16` is problematic unless there is a guarantee that the value will not be greater than `i16::MAX`. Where is this checked?

joshleeb: Added the check in

seanmonstar:
Thanks!

PR #258, opened 2018-10-05
impl Display for Version

blackgear:
`http::Version` could impl `Display` just like `Uri` and `Method`, so that we can use a much consistent way during `format!`

similar code ref: https://github.com/hyperium/http/blob/master/src/uri/mod.rs#L969-L993

### DIFF
--- a/src/version.rs
+++ b/src/version.rs
@@ -54,7 +54,7 @@ impl Default for Version {
     }
 }
 
-impl fmt::Debug for Version {
+impl fmt::Display for Version {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         use self::Http::*;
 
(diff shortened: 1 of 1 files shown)


seanmonstar:
For reference, the reason it didn't implement `Display` was because it doesn't necessarily make sense. The context that you would need to do this is likely to encode the version to a transport, but depending on the version used, the output differs.

Or rather, the output of `HTTP/1.x` only makes sense for HTTP/1.0 and 1.1, but not for HTTP2. In the context of HTTP2, the version isn't even encoded at all.

Now, I don't know if that is a strong enough argument to not implement `Display`, and could be convinced it's useful, but that's the reason so far (if it's decided that it still shouldn't implement `Display`, perhaps some comments should be added in the code so it's clear it isn't an oversight).

@carllerche what do you think?

PR #265, opened 2018-10-18
change Port<'a> to be Port<Repr>

### LINKED ISSUE #263: [v0.1.14 blocker] Should Port<'a> have a different name?
In #173, which is now in master, we added a `Port<'a>`, which includes not only the `u16` port number, but also the `&str` of the port. This allows efficiently asking for the "port" part of a `Uri` as a string, without needing to get a `u16` and allocate a *new* `String` by calling `port_num.to_string()`.

In #219, also in master, a `uri::Builder` was introduced. The idea is to allow for easily building up a `Uri` from fallible parts. It seems like it could be nice to add a `port` setter, which takes a generic `P where Port: HttpTryFrom<P>`. This would allow a user to pass a `u16`, a `&str`, or a `Port` from a different `Uri`.

However, with `Port<'a>`, adding this API to `uri::Builder` would be impossible. So, should `Port<'a>` be renamed to something else? It'd be useful to find other precedents, but until they are found and listed, if renaming is desired, some possible names are:

- `PortRef<'a>`
- `PortSlice<'a>`
- `PortPart<'a>`

As this type isn't released yet (only on master), and it affects whether we can add to the `uri::Builder` API, I believe finding an answer should block releasing v0.1.14.

seanmonstar:
Closes #263

### DIFF
--- a/src/uri/port.rs
+++ b/src/uri/port.rs
@@ -1,21 +1,14 @@
-use std::{fmt, num};
-use std::str::FromStr;
+use std::fmt;
+
+use super::{ErrorKind, InvalidUri};
 
 /// The port component of a URI.
-#[derive(Debug)]
-pub struct Port<'a> {
-    bytes: &'a str,
+pub struct Port<T> {
     port: u16,
+    repr: T,
 }
 
-impl<'a> Port<'a> {
-    /// Converts a `str` to a port number.
-    ///
-    /// The supplied `str` must be a valid u16.
-    pub(crate) fn from_str(bytes: &'a str) -> Result<Self, num::ParseIntError> {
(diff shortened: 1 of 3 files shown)


src/uri/port.rs
```diff
@@ -46,41 +60,58 @@ impl<'a> Port<'a> {
...
     /// assert_eq!(port.as_str(), "80");
     /// ```
-    pub fn as_str(&self) -> &'a str {
-        self.bytes
+    pub fn as_str(&self) -> &str {
+        self.repr.as_ref()
+    }
+}
+
+impl<T> fmt::Debug for Port<T> {
```
carllerche: For forwards compat, do we want to bound `T: fmt::Debug` even if it is not included?

PR #266, opened 2018-10-19
add debug assertions for unchecked constructors

seanmonstar:


src/header/value.rs
```diff
@@ -173,9 +173,18 @@ impl HeaderValue {
     /// within the buffer.
     #[inline]
     pub unsafe fn from_shared_unchecked(src: Bytes) -> HeaderValue {
-        HeaderValue {
-            inner: src,
-            is_sensitive: false,
+        if cfg!(debug_assertions) {
+            match HeaderValue::from_shared(src.clone()) {
```
carllerche: I would avoid the clone here. If we want to include the invalid bytes in the error message, it should probably be part of the returned `err`.

Cloning could have additional unexpected effect, like promoting `Vec<u8>` -> `Arc<Vec<u8>>`.

PR #268, opened 2018-10-29
Adding the `QueryItem` struct in path

yageek:
As requested in #206, building an URI from a builder should have a smaller granularity.
A struct `QueryItem` has been added. It holds the elements of a query component.
A method `new` inside `PathAndQuery` has been added. It  that requires a `path` and a list of `QueryItem` and build a new `PathAndQuery` from them.

The URI builder could hold both a `String` and a `Vec<QueryItem>` fields instead of one `parts` field and could provide new methods to configure the path and the query separately.

### DIFF
--- a/src/uri/path.rs
+++ b/src/uri/path.rs
@@ -1,11 +1,20 @@
-use std::{cmp, fmt, str};
 use std::str::FromStr;
+use std::{cmp, fmt, str};
 
 use bytes::Bytes;
 
+use super::{ErrorKind, InvalidUri, InvalidUriBytes, URI_CHARS};
 use byte_str::ByteStr;
 use convert::HttpTryFrom;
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks @yageek for thinking about this! I think we should discuss a bit more what the API should look like.

- Assuming that this should be an addition to the `Builder`, we should consider what new method(s) should be added.
- I'd suggest that this shouldn't (yet) add `PathAndQuery::new`, as that's a pretty coveted constructor name. We should start with just the `Builder`.
- It's preferred that any public `struct` have its fields be private, so that internal representations can be changed without worry of breaking changes.

(If you'd rather discuss these on the tracking issue #206, that's fine too!)

yageek:
I think you're right, I may attack the problem in the wrong order. I'll put some comments on #206

PR #269, opened 2018-10-29
Allow the percent character in the userinfo of an authority.

arsing:
The username and password in an authority are percent-encoded.


seanmonstar:
Seems fine, mind including a comment next to the code with a link to the relevant RFC/section so Future Us doesn't forget the *why*?

arsing:
Do you have any preference between https://tools.ietf.org/html/rfc3986#section-3.2.1 and https://url.spec.whatwg.org/#set-the-username ?

seanmonstar:
Thanks!

PR #271, opened 2018-11-01
Support more tokens for header names

Eijebong:
According to the RFC both `^` and ` are valid in a header name.
This is also making some wpt test fail in servo since they're trying to
build a headername with all those characters

### DIFF
--- a/src/header/name.rs
+++ b/src/header/name.rs
@@ -973,11 +973,15 @@ standard_headers! {
 ///
 /// ```not_rust
 ///       field-name     = token
-///       token          = 1*<any CHAR except CTLs or separators>
 ///       separators     = "(" | ")" | "<" | ">" | "@"
 ///                      | "," | ";" | ":" | "\" | <">
 ///                      | "/" | "[" | "]" | "?" | "="
 ///                      | "{" | "}" | SP | HT
+///       token          = 1*tchar
+///       tchar          = "!" / "#" / "$" / "%" / "&" / "'" / "*"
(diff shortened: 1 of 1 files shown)


src/header/name.rs
```diff
@@ -990,7 +994,7 @@ const HEADER_CHARS: [u8; 256] = [
         0,     0,     0,     0,     0,  b'a',  b'b',  b'c',  b'd',  b'e', //  6x
      b'f',  b'g',  b'h',  b'i',  b'j',  b'k',  b'l',  b'm',  b'n',  b'o', //  7x
      b'p',  b'q',  b'r',  b's',  b't',  b'u',  b'v',  b'w',  b'x',  b'y', //  8x
-     b'z',     0,     0,     0,     0,  b'_',     0,  b'a',  b'b',  b'c', //  9x
```
seanmonstar: It looks like this removes the `z`?

Eijebong: Oh woups... I copy/pasted it from the HTTP2 tokens

Cargo.toml
```diff
@@ -4,7 +4,7 @@ name = "http"
 # - Update html_root_url in lib.rs.
 # - Update CHANGELOG.md.
 # - Create git tag
-version = "0.1.13"
+version = "0.1.14"
```
seanmonstar:
```suggestion
version = "0.1.13"
```

Mind leaving the version change out? We can bump it as part of the changelog generation commit.

Eijebong: Oh sure :)

PR #276, opened 2018-11-05
Add path_segments to PathAndSegments

phillyai:
Add path_segments to PathAndSegments and add test_uri_parse_segments to tests. 
And I just got noticed rustfmt auto formatted two code files.

### DIFF
--- a/src/uri/path.rs
+++ b/src/uri/path.rs
@@ -1,11 +1,11 @@
-use std::{cmp, fmt, str};
 use std::str::FromStr;
+use std::{cmp, fmt, str};
 
 use bytes::Bytes;
 
+use super::{ErrorKind, InvalidUri, InvalidUriBytes, URI_CHARS};
 use byte_str::ByteStr;
 use convert::HttpTryFrom;
-use super::{ErrorKind, InvalidUri, InvalidUriBytes, URI_CHARS};
 
 /// Represents the path component of a URI
 #[derive(Clone)]
@@ -56,10 +56,7 @@ impl PathAndQuery {
                     //
                     // Allowed: 0x21 / 0x24 - 0x3B / 0x3D / 0x3F - 0x7E
                     match b {
-                        0x21 |
-                        0x24...0x3B |
-                        0x3D |
-                        0x3F...0x7E => {},
+                        0x21 | 0x24...0x3B | 0x3D | 0x3F...0x7E => {}
                         _ => return Err(ErrorKind::InvalidUriChar.into()),
                     }
                 }
@@ -108,8 +105,7 @@ impl PathAndQuery {
     pub fn from_static(src: &'static str) -> Self {
         let src = Bytes::from_static(src.as_bytes());
 
-        PathAndQuery::from_shared(src)
-            .unwrap()
+        PathAndQuery::from_shared(src).unwrap()
     }
 
     pub(super) fn empty() -> Self {
@@ -168,6 +164,27 @@ impl PathAndQuery {
         }
 
         ret
+    }
+    /// Returs the segments of path component
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use http::uri::*;
+    ///
+    /// let path_and_query: PathAndQuery = "/hello/world".parse().unwrap();
+    ///
+    /// assert!(path_and_query.path_segments().is_some());
+    /// assert_eq!(path_and_query.path_segments().unwrap().collect::<Vec<_>>(), vec!["hello",
+    /// "world"])
+    /// ```
+    pub fn path_segments(&self) -> Option<str::Split<char>> {
+        let path = self.path();
+        if path.starts_with('/') {
+            Some(path[1..].split('/'))
+        } else {
+            None
+        }
     }
 
     /// Returns the query string component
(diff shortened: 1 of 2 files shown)


src/uri/path.rs
```diff
@@ -1,11 +1,11 @@
-use std::{cmp, fmt, str};
 use std::str::FromStr;
+use std::{cmp, fmt, str};
 
 use bytes::Bytes;
 
+use super::{ErrorKind, InvalidUri, InvalidUriBytes, URI_CHARS};
 use byte_str::ByteStr;
 use convert::HttpTryFrom;
-use super::{ErrorKind, InvalidUri, InvalidUriBytes, URI_CHARS};
```
carllerche: Could this and the line above be reverted?

src/uri/path.rs
```diff
@@ -56,10 +56,7 @@ impl PathAndQuery {
                     //
                     // Allowed: 0x21 / 0x24 - 0x3B / 0x3D / 0x3F - 0x7E
                     match b {
-                        0x21 |
-                        0x24...0x3B |
-                        0x3D |
-                        0x3F...0x7E => {},
+                        0x21 | 0x24...0x3B | 0x3D | 0x3F...0x7E => {}
```
carllerche: Could this be reverted?

src/uri/path.rs
```diff
@@ -108,8 +105,7 @@ impl PathAndQuery {
     pub fn from_static(src: &'static str) -> Self {
         let src = Bytes::from_static(src.as_bytes());
 
-        PathAndQuery::from_shared(src)
-            .unwrap()
+        PathAndQuery::from_shared(src).unwrap()
```
carllerche: Could this be reverted?

src/uri/path.rs
```diff
@@ -169,6 +165,27 @@ impl PathAndQuery {
...
+    /// ```
+    /// # use http::uri::*;
+    ///
+    /// let path_and_query: PathAndQuery = "/hello/world".parse().unwrap();
+    ///
+    /// assert!(path_and_query.path_segments().is_some());
+    /// assert_eq!(path_and_query.path_segments().unwrap().collect::<Vec<_>>(), vec!["hello",
+    /// "world"])
+    /// ```
+    pub fn path_segments(&self) -> Option<str::Split<char>> {
```
carllerche: This should probably return a custom iterator: `struct PathSegments<str::Split<...>>`

Is `Option` significant in this case or can the iterator be returned directly? `Uri::path_and_query` already returns an `Option`.

src/uri/tests.rs
```diff
@@ -1,6 +1,6 @@
 use std::str::FromStr;
 
-use super::{ErrorKind, InvalidUri, Uri, URI_CHARS, Port};
+use super::{ErrorKind, InvalidUri, Port, Uri, URI_CHARS};
```
carllerche: Could this be reverted?

src/uri/tests.rs
```diff
@@ -12,9 +12,9 @@ fn test_char_table() {
 }
 
 macro_rules! part {
-    ($s:expr) => (
+    ($s:expr) => {
```
carllerche: Could this be reverted?

phillyai:
All right, First, I reverted unrelated changes.

src/uri/path.rs
phillyai: Done!

src/uri/path.rs
phillyai: You're right, `Option` in this context is meanless.

phillyai: And custom iterator you said, I need help.. I'm so new to rust, When I write any custom iterator, compiler raise lifetime error. T^T

src/uri/path.rs
```diff
@@ -170,6 +170,24 @@ impl PathAndQuery {
...
+    /// # use http::uri::*;
+    ///
+    /// let path_and_query : PathAndQuery = "/hello/world".parse().unwrap();
+    ///
+    /// assert_eq!(
+    ///     path_and_query.path_segments().collect::<Vec<_>>(),
+    ///     vec!["hello", "world"]
+    /// );
+    /// ```
+    pub fn path_segments(&self) -> str::Split<char> {
```
carllerche: Could you return a new type here? Some type that wraps `str::Split<char>` and implements `Iterator`:

```rust
struct PathSegments(str::Split<char>);
```

This is for forwards compatibility as it allows us to change the implementation details of the iterator.

I'm also on the fence about whether it should be `path_segments` or just `segments`. It is a method on the `PathAndQuery` type, so it could go either way...

phillyai: Sorry to bother you, but I created one. Can you check?

src/uri/path.rs
```diff
@@ -7,6 +7,18 @@ use byte_str::ByteStr;
 use convert::HttpTryFrom;
 use super::{ErrorKind, InvalidUri, InvalidUriBytes, URI_CHARS};
 
+/// Represents the segments of a URI
+#[derive(Debug)]
+pub struct PathSegments<'a>(str::Split<'a, char>);
+
+impl<'a> std::ops::Deref for PathSegments<'a> {
```
carllerche: Instead of implementing `Deref`, could you implement, could we implement `Iterator<Item = &'a str>` and `DoubleEndedIterator`? The impl would proxy to `str::Slice<'a, char>`'s implementation.

carllerche: I also think that the the CI failure is due to referencing `std` instead of `::std`.

PR #280, opened 2018-11-28
Add Clone to Request

TheNeikos:
I'm currently using `hyper` and it exposes the Request through the `http` crate here. 

One thing I am currently needing is to `clone` a Request, this would implement the http part of it.

Since `Extensions` uses an `AnyMap` this causes some problems, as a clone is not a complete but a partial clone. I have added a part about documentation here. However, as Extensions seem to be used for 'Only for this Extensions' semantics, this could be a good thing.

### DIFF
--- a/src/extensions.rs
+++ b/src/extensions.rs
@@ -167,6 +167,12 @@ impl Extensions {
     }
 }
 
+impl Clone for Extensions {
+    fn clone(&self) -> Extensions {
+        Extensions::new()
+    }
+}
+
 impl fmt::Debug for Extensions {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         f.debug_struct("Extensions")
--- a/src/request.rs
+++ b/src/request.rs
@@ -153,6 +153,7 @@ use version::Version;
 /// #
 /// # fn main() {}
 /// ```
+#[derive(Clone)]
 pub struct Request<T> {
     head: Parts,
     body: T,
@@ -162,6 +163,7 @@ pub struct Request<T> {
 ///
 /// The HTTP request head consists of a method, uri, version, and a set of
 /// header fields.
+#[derive(Clone)]
 pub struct Parts {
     /// The request's method
     pub method: Method,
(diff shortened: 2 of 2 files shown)


TheNeikos:
Although, I'd like to add, I find it odd that `Request` has an `Extensions` field. If `http` is to be a common crate for type definitions then this seems at odd as it is quite an opinionated feature.

I have read the previous discussion here: https://github.com/hyperium/http/pull/40 but I am not sure if `http` is the place for this functionality... but this is orthogonal to this PR anyway :grinning:

src/extensions.rs
```diff
@@ -167,6 +167,12 @@ impl Extensions {
     }
 }
 
+impl Clone for Extensions {
```
sfackler: A clone implementation that doesn't actually produce a copy seems not great...

TheNeikos:
Well, it is not optimal I agree, this can be closed of course if its not wanted.

Do you think there could be an alternative for this? Maybe allow to clone the body without cloning the whole request?

seanmonstar:
> Maybe allow to clone the body without cloning the whole request?

This should already be possible, right?

```rust
let req = Request::new("hello");
let body2 = req.body().clone();
```

TheNeikos:
Yes, but you would have to do that individually for all the parts. 
I guess that would be a workaround if the extensions map is to stay.

hamza1311:
> Yea, as long as `Extensions` cannot be cloned, I'm not so sure we can just implement `Clone` here.

@seanmonstar, is it fine to implement `Clone` here if `Extensions` can be cloned? If so, we can make extensions use [`CloneAny`](https://docs.rs/anymap2/0.13.0/anymap2/any/trait.CloneAny.html) instead of [`Any`](https://docs.rs/anymap2/0.13.0/anymap2/any/trait.Any.html). This would require using `anymap2` instead of `anymap` but that is an improvement considering `anymap` is currently seems to be unmaintained.

I'd fine with working on it, just want to make sure if it's something that should be incorporated.

seanmonstar:
Requiring extensions to be `Clone` could be hard. It would make it hard to store unique things for a request, something where you only expect there to be one of, and that perhaps later you take out and mutate or consume. I'd suggest opening a new issue to discuss that more if you want.

hamza1311:
> Requiring extensions to be `Clone` could be hard. It would make it hard to store unique things for a request, something where you only expect there to be one of, and that perhaps later you take out and mutate or consume.

Makes sense. It seems that it is fine as-is. If we made extensions `Clone`, it would mean only types that implement `Clone` could be used, which might not be ideal.

PR #281, opened 2018-12-02
Add functions to split URI query parts

landesfeind:
Hi,
this PR adds a function to parse the URI query into a HashMap to allow easier retrieval of the values plus some convenience functions. Finally, it is quite easy to just do:
```
let my_query_value = match uri.query_param_first("foo") {
   Some(value) => value,
   None => "my-default-value"
}
```
The code is provided as is and I hope it helps someone. Please be aware that I am relatively new to Rust and hope that the added tests ensure that I did not screw up totally ;-)
Best,
Manuel

### DIFF
--- a/src/uri/path.rs
+++ b/src/uri/path.rs
@@ -256,14 +256,14 @@ impl PathAndQuery {
         r
     }
 
-    /// Parses the query string component and return the first value to the given key
+    /// Parses the query string component and return the values to the given key
     ///
     /// # Examples
     ///
     /// ```
     /// # use http::uri::*;
     /// let path_and_query: PathAndQuery = "/hello/world?key=value&foo=bar".parse().unwrap();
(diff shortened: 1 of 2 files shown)


carllerche:
Thanks for the work and posting this.

I agree that it would be nice to have a simple API for getting query parameters out. One difficulty is that `http` is attempting to be as opinionated as possible and there is no unopinionated way to parse query parameters. You'd *probably* want to do something like what the browser does, but how do you handle nested params, etc...

Given this, I'm not sure if this code should live in the `http` crate vs. a separate crate. Thoughts @seanmonstar ?

seanmonstar:
> My view is that there are too many views on how to parse query params for anything to be included in this crate.

I agree. There's different strategies even for handling "arrays" and "maps" in query params.

Thank you for the PR! At this point, I'm going to close and suggest people use one of the mention serde parsers.

Issue #282, opened 2018-12-06
get_method, get_uri on request::Builder


kazuk:
I am writing Azure Storage REST Client and implementing [Authorize with Shared Key](https://docs.microsoft.com/en-us/rest/api/storageservices/authorize-with-shared-key). To create signatures, I need HTTP Method(VERB) , header values, and request URI on the current request. (See  [Blob, Queue, and File Services (Shared Key authorization)](https://docs.microsoft.com/en-us/rest/api/storageservices/authorize-with-shared-key#blob-queue-and-file-services-shared-key-authorization))

I wish to add functions to `request::Builder` 

- `get_method`
- `get_uri`  
- `get_headers`

then I can implement sign_request as bellow
```
fn sign_request(&self, request_builder: Builder ) -> Builder {
    // build string_to_sign
    let string_to_sign =format!( "{}\n.....", request_builder.get_method(). ... );
    // compute signature 
    let signature = ....;
    // add Authorization header to Builder and return it.
    request_builder.header( AUTHORIZATION, "SharedKey {}:{}", self.account_name,signature )
}
```

Can you accept if I send changes?

Or is there any security reason that `request::Builder` shouldn't provide HTTP method etc?

seanmonstar:
On one hand, it certainly *seems* reasonable to want to view what fields have already been set. On the other, I can't remember any *other* builder patterns where getters are included...

kazuk:
https://en.wikipedia.org/wiki/Builder_pattern

C# builder having getter.  because C# property MUST have getter.

C++,Java,etc builder not having getter. 
But user can view current value from builder and update as bellow.

```
let temp_order = builder.build();
let tax = compute_tax( temp_order.item_lines() );
let taxed_order = builder.set_tax(tax).build();
```

in Rust, Builder consumed on first line build() and ends life. last line build() to make compilation error.

- from 'C# Builder' example, Builder have getter when language requirement or limitation hits situation.
- in Rust, Builder causes some of compilation errors, user do 'view current value'. 

IMHO
 'some of compilation errors' as 'language requirement or limitation hits situation' , and then Builder can have getter.

carllerche:
I've wanted this every so often. I would be fine with having `get_*` functions.

carllerche:
Though, if we are to bikeshed a bit, generally Rust doesn't have `get_*` prefix. We obviously can't have no prefix since that fn is used by the setter...

I'm leaning towards just adding `get_` prefixes, but I'm open to other thoughts.

[...]

PR #284, opened 2018-12-13
Add method_ref, uri_ref, headers_ref to request::Builder

kazuk:
implementation for https://github.com/hyperium/http/issues/282

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -771,6 +771,34 @@ impl Builder {
         self
     }
 
+    /// Get the HTTP Method for this request.
+    /// 
+    /// By default this is `GET`.
+    /// if builder has error, returns None.
+    /// 
+    /// # Examples
+    /// 
+    /// ```
+    /// # use http::*;
+    /// 
+    /// let mut req = Request::builder();
+    /// assert_eq!(req.get_method(),Some(Method::GET));
+    /// req.method("POST");
+    /// assert_eq!(req.get_method(),Some(Method::POST));
+    /// req.method("DELETE");
+    /// assert_eq!(req.get_method(),Some(Method::DELETE));
+    /// ```
+    pub fn get_method(&self) -> Option<Method>
+    {
+        if self.err.is_some() {
+            return None
+        }
+        match &self.head {
+            Some(head) => Some(head.method.clone()),
+            None => None
+        }
+    }
+
     /// Set the URI for this request.
     ///
     /// This function will configure the URI of the `Request` that will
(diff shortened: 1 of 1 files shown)


src/request.rs
```diff
@@ -771,6 +771,34 @@ impl Builder {
...
+    /// # use http::*;
+    /// 
+    /// let mut req = Request::builder();
+    /// assert_eq!(req.get_method(),Some(Method::GET));
+    /// req.method("POST");
+    /// assert_eq!(req.get_method(),Some(Method::POST));
+    /// req.method("DELETE");
+    /// assert_eq!(req.get_method(),Some(Method::DELETE));
+    /// ```
+    pub fn get_method(&self) -> Option<Method>
```
carllerche: I'd probably return a reference here.

src/request.rs
```diff
@@ -860,6 +913,33 @@ impl Builder {
...
+    /// # use http::header::HeaderValue;
+    /// # use http::request::Builder;
+    /// let mut req = Request::builder();
+    /// req.header("Accept", "text/html")
+    ///    .header("X-Custom-Foo", "bar");
+    /// let headers = req.get_headers().unwrap();
+    /// assert_eq!( headers["Accept"], "text/html" );
+    /// assert_eq!( headers["X-Custom-Foo"], "bar" );
+    /// ```
+    pub fn get_headers(&self ) -> Option<HeaderMap<HeaderValue>> {
```
carllerche: Extra space.

src/request.rs
```diff
@@ -771,6 +771,34 @@ impl Builder {
...
+    /// assert_eq!(req.get_method(),Some(Method::POST));
+    /// req.method("DELETE");
+    /// assert_eq!(req.get_method(),Some(Method::DELETE));
+    /// ```
+    pub fn get_method(&self) -> Option<Method>
+    {
+        if self.err.is_some() {
+            return None
+        }
+        match &self.head {
```
seanmonstar: You should be able to remove most of these `&`s:

```rust
match self.head {
    Some(ref head) => Some(&head.method),
    None => None,
}
```

src/request.rs
```diff
@@ -800,6 +828,31 @@ impl Builder {
...
+    /// 
+    /// ```
+    /// # use http::*;
+    /// 
+    /// let mut req = Request::builder();
+    /// assert_eq!(req.get_uri().unwrap().to_string(), "/" );
+    /// req.uri("https://www.rust-lang.org/");
+    /// assert_eq!(req.get_uri().unwrap().to_string(), "https://www.rust-lang.org/" );
+    /// ```
+    pub fn get_uri(&self) -> Option<Uri>
```
seanmonstar: This should return an `Option<&Uri>`. The user can decide if they need to clone it, but if not, no need to be wasteful :)

src/request.rs
seanmonstar: Should return `Option<&HeaderMap>`

PR #286, opened 2019-01-01
Introduce IntoRequest and IntoResponse traits

davidbarsky:
(I'm sorry for the the long writeup. While the code changes themselves are small, this introduces a new public interface to `hyperium/http` and, if this change is accepted, subsequent PRs to Hyper et. al.)

These changes are kinda small, but they're a step in a direction allows for both more easily retryable requests and be strongly typed. I'll expand on each.

**Retries**

To retry requests, the request + payload must implement `Clone` or have _some_ mechanism of cloning. For instance, in Linkerd2, retries are handled via an [internal `TryClone` trait](https://github.com/linkerd/linkerd2-proxy/blob/2f2050537dd8caeaf5704e66a3fd7be89f2455a3/src/app/dst.rs#L84). While _most_ fields in `Request`/`Response` implement clone, `Extensions` can't easily implement clone (e.g., #40, #280). Additionally, not all of Hyper's Body's kinds can easily be [retried or even cloned](https://github.com/hyperium/hyper/blob/master/src/body/body.rs#L30).

Short of implementing `Clone` on `Extensions` and refactoring Hyper's `Body` to support retryable payloads—both non-trivial and likely breaking changes—I think that the next best option would be to provide a mechanism to convert arbitrary types into Requests and Responses. In the "allow retries" use case, an author can write several types that implement `Clone`, and Hyper's Client (pending this PR's approval, I'd be happy to open a PR on Hyper) would accept any type implementing `IntoRequest<B: hyper::body::Payload>`.

**Strongly Typed Requests/Responses**

This PR took inspiration from `std::iterator::IntoIterator`. A quick glance at [tower-web](https://github.com/carllerche/tower-web/blob/master/src/response/response.rs) makes me think that this PR might provide a more general approach for responses, and allows people vending interfaces/libraries to HTTP services to provide typed interfaces more easily.


davidbarsky:
I'm thinking that this should also entail corresponding `FromRequest<B>` and `FromResponse<B>` traits as well as some sort of generalizable extractor.

carllerche:
I think this is an interesting path to explore, but probably not in the core crate. Similar to typed headers, this should first be explored in a separate crate.

Some thoughts, I'm not sure how this supports retries given that `IntoRequest` takes `self`. I also think that you can't require *all* requests are retry-able, which is why linkerd uses a `TryClone` trait.

Also, the body type should probably be an associated type and not a generic.

davidbarsky:
> I think this is an interesting path to explore, but probably not in the core crate. Similar to typed headers, this should first be explored in a separate crate.

Fair enough. I can create a new crate?

> Some thoughts, I'm not sure how this supports retries given that IntoRequest takes self. 

The idea is that the source type would implement `Clone`, so that the source value would be cloned _before_ `into_request` is called. I now realize that this would require a clone before converting into request for every retryable request, which is unnecessarily expensive. Instead, the request should probably be cloned on retry.

> I also think that you can't require all requests are retry-able, which is why linkerd uses a TryClone trait.

I agree, yes. My thinking is that the _source value_—not the the request itself—can be cloned & retried upon conversion to a request. This PR attempts to provide that mechanism.

carllerche:
Would there be any built-in implementations of `IntoRequest` / `IntoResponse` for std types? I'm not sure if there are any obvious implementations.

davidbarsky:
I don't think so. The more I think about this, the more I think this functionality should be handled in Tower as per Sean's [gist](https://gist.github.com/seanmonstar/2469a8e544ada2d31d13c8ee54f17d48). I think I'll close this issue in the meantime.

PR #289, opened 2019-01-11
Running cargo fmt

seanmonstar:


benches/header_map/basic.rs
```diff
@@ -2,13 +2,13 @@ macro_rules! bench {
...
-            use seahash::SeaHasher;
+            use super::custom_hdr;
             use fnv::FnvHasher;
-            use std::hash::BuildHasherDefault;
             use http::header::*;
+            use seahash::SeaHasher;
+            use std::hash::BuildHasherDefault;
             #[allow(unused_imports)]
-            use super::custom_hdr;
+            use test::{self, Bencher};
```
carllerche: I thought rustfmt didn't touch macros?

src/header/map.rs
```diff
@@ -1066,20 +1075,27 @@ impl<T> HeaderMap<T> {
...
-        where K: Hash + Into<HeaderName>,
-              HeaderName: PartialEq<K>,
+    where
+        K: Hash + Into<HeaderName>,
+        HeaderName: PartialEq<K>,
     {
         self.reserve_one();
 
         insert_phase_one!(
-            self, key, probe, pos, hash, danger,
```
carllerche: This change is annoying :( I thought rustfmt didn't attempt to format macros.

src/header/mod.rs
```diff
@@ -74,118 +74,34 @@ mod map;
 mod name;
 mod value;
 
+#[rustfmt::skip]
```
carllerche: I'd remove

src/status.rs
```diff
@@ -531,33 +532,31 @@ macro_rules! status_code_strs {
 }
 
 status_code_strs!(
-    100, 101, 102, 103, 104, 105, 106, 107, 108, 109, 110, 111, 112, 113, 114, 115, 116, 117, 118, 119,
```
carllerche: This is a terrible change and should probably be skipped.

carllerche:
Thanks for doing this.

There are a few weird transformations. Only one seems like it should be a "bug" (in `if / else` statements, the `{` is moved to the next line, but nothing else in the condition is broken up).

That said, the `http` crate is unique in that a number of imports and other code snippets are formatted in a unique way for legibility. `rustfmt::skip` exists for these cases, but there is a bug with imports: https://github.com/rust-lang/rustfmt/issues/3273.

For `http`, I think we should hold off until https://github.com/rust-lang/rustfmt/issues/3273 is fixed.

PR #290, opened 2019-01-21
Add len() to Extensions

davidbarsky:
A convenience method that makes Extensions more HashMap-like.

### DIFF
--- a/src/extensions.rs
+++ b/src/extensions.rs
@@ -147,6 +147,22 @@ impl Extensions {
             })
     }
 
+    /// Returns the number of entries in `Extensions`.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use http::Extensions;
+    /// let mut ext = Extensions::new();
+    /// ext.insert(5i32);
+    /// assert_eq!(ext.len(), 1);
+    /// ext.remove::<i32>();
+    /// assert_eq!(ext.len(), 0);
+    /// ```
+    pub fn len(&self) -> usize {
+        self.map.as_ref().map_or(0, |m| m.len())
+    }
+
     /// Clear the `Extensions` of all inserted extensions.
     ///
     /// # Example


seanmonstar:
Just curious, what use-case are you thinking of where you'd want to utilize this?

davidbarsky:
> Just curious, what use-case are you thinking of where you'd want to utilize this?

To work around [Extensions](https://docs.rs/http/0.1.14/http/struct.Extensions.html) not implementing `Clone`, which is being is done as part of the work in: https://github.com/tower-rs/tower-hyper.

The alternative to this would be make Extensions implement clone. One candidate for this would to borrow Tide's [implementation](https://github.com/rust-net-web/tide/blob/master/src/configuration/mod.rs#L15), but I worry this would be a breaking change.

davidbarsky:
> What would one do with knowing the length of the extensions?

Mainly, if it's safe to do a partial clone on a Request, especially as part of part of Tower Retry's `try_clone()`. A conservative option would be to _not_ clone the request if the extensions map has any values populated. I'm not sure if this is the right way, but I'd like to be able to express a `try_clone()` across a lot of HTTP/1 libraries in some sort of generic/composable way.

(Perhaps I'm misunderstanding what Extensions is used for in practice—I'd like to do this safely and sanely.)

> I know in Linkerd2 I've stored things in extensions that definitely couldn't be clone (well, I guess anything can, but it'd make no sense... It'd end up being an Arc<Mutex<Option<..>>>...).

Yep, hence my concern about requiring a `Clone` bound on Extension.

seanmonstar:
> A conservative option would be to not clone the request if the extensions map has any values populated.

That's reasonable, perhaps as a default clone policy. What if there was just `Extensions::is_empty()`?

> (Perhaps I'm misunderstanding what Extensions is used for in practice—I'd like to do this safely and sanely.)

I've seen them used for various things:

- To attach a `Session` to a request, after going through some authentication middleware.
- Adding a marker or label that is used at the end when reporting metrics.
- After checking out a DB connection, saving it with the request.

davidbarsky:
I thought I replied, sorry about that.

> That's reasonable, perhaps as a default clone policy. What if there was just Extensions::is_empty()?

That can work!

> I've seen them used for various things:
> - To attach a Session to a request, after going through some authentication middleware.
> - Adding a marker or label that is used at the end when reporting metrics.
> - After checking out a DB connection, saving it with the request.

I wonder if it makes sense to move this functionality into Tower itself—it seems to be taking role similar to Go's `context.Context` minus the cancelation.

davidbarsky:
I should clarify. When I said:

> I wonder if it makes sense to move this functionality into Tower itself—it seems to be taking role similar to Go's context.Context minus the cancelation.

...I meant at some point in the future, with a breaking semver bump.

seanmonstar:
I think the outcome here is to leave as-is for now. Let me know if I'm wrong :)

PR #294, opened 2019-02-04
Add awareness of HTTP/3

LPardue:
The HTTP mapping to the QUIC transport was renamed to HTTP/3 in [Internet Draft version 17](https://tools.ietf.org/html/draft-ietf-quic-http-17).

This PR just adds the basics (enum, string) that acknowledge that HTTP/3 is now the official name.

### DIFF
--- a/src/header/name.rs
+++ b/src/header/name.rs
@@ -1013,6 +1013,7 @@ const HEADER_CHARS: [u8; 256] = [
         0,     0,     0,     0,     0,     0                              // 25x
 ];
 
+/// Valid header name characters for HTTP/2.0 and HTTP/3.0
 const HEADER_CHARS_H2: [u8; 256] = [
     //  0      1      2      3      4      5      6      7      8      9
         0,     0,     0,     0,     0,     0,     0,     0,     0,     0, //   x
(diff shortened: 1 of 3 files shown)


seanmonstar:
Thanks for this, I really want to proceed here, but just calling out that thanks to (in my opinion, a compiler bug) the way the compiler leaks exhaustive-ness of the wrapped enum, adding this new constant is technically a breaking change. See #155.

LPardue:
Ah shucks. Thanks for the link.

carllerche:
Because we (foolishly apparently) assumed HTTP/3 would not be coming soon...

seanmonstar:
@carllerche we didn't think it'd come so soon, but we *did* try to design the `Version` type to be extendable, by making it an opaque `struct`. However, the compiler "leaks" that you can exhaustively match all the constants.

carllerche:
Ah yes...

quininer:
HTTP/3 is not far off, can we start building a `0.2.0-alpha` branch so the h3 implementation can using it?

carllerche:
We can probably try to coordinate a release w/ `Future` reaching std stable.

RReverser:
Can H3 be added under an optional feature for now with `#[cfg(feature = "h3")]` guards on everything? Then it won't break downstream crates that didn't opt-in to the feature yet.

seanmonstar:
@RReverser unfortunately not. Features are additive. If one dependency didn't exprct the extra variant, whereas another turned on the feature, it'd still break compilation.

RReverser:
*sigh* Right, that reduces the risk but doesn't eliminate it...

djc:
Can we add this under an off-by-default feature for now? That way, we can allow experimentation with H3 support across the ecosystem without waiting for the next API-breaking release.

djc:
Ah, sorry -- I see this was discussed above. Still, if we want to allow HTTP 3 experiments across the ecosystem (which would be nice given the H3 timeline -- handed over to the steering group for review in July) it seems we need to unblock something here soonish.

seanmonstar:
I've started a [0.2.x](https://github.com/hyperium/http/tree/0.2.x) branch, which we can start merging these kinds of things into.

PR #296, opened 2019-02-18
Permit more characters in URI paths

seanmonstar:
The https://url.spec.whatwg.org/#path-state allows more characters than we currently allow. This was noticed in https://github.com/seanmonstar/reqwest/issues/442.


sfackler:
We just ran into this as well! Could you cut a release with these changes?

Issue #300, opened 2019-03-08
Parse interprets protocol relative url as relative path url


Insomniak47:
Parsing of a protocol relative uri string doesn't produce a uri that contains the proper elements in the proper members.

cargo version: 1.33.0 (f099fe94b 2019-02-12)
http version: http = "0.1.16"

Example:
```
    let a = "//en.wikipedia.org/interesting_article".parse::<Uri>().unwrap();
    println!("{:#?}",a);
    println!("Scheme:    {:#?}", a.scheme_part().map_or_else(|| "", |x| x.as_str()));
    println!("Authority: {:#?}", a.authority_part().map_or_else(|| "", |x| x.as_str()));
    println!("Path:      {:#?}", a.path_and_query().map_or_else(|| "", |x| x.as_str()));
```

Expected Output:
```
//en.wikipedia.org/interesting_article
Scheme:    ""
Authority: "en.wikipedia.org"
Path:      "/interesting_article"
```

Actual Output
```
//en.wikipedia.org/interesting_article
Scheme:    ""
Authority: ""
Path:      "//en.wikipedia.org/interesting_article"
```

bluetech:
I don't think this form is allowed in an HTTP [request target](https://tools.ietf.org/html/rfc7230#section-5.3), which is what `http::Uri` represents according to its docstring. You can double-check me by following the EBNF from this link, and down to [RFC3986](https://tools.ietf.org/html/rfc3986). `absolute-form` only allows an `absolute-URI` which does not allow a relative scheme.

If I got it right, I think this change should be reverted, to save people from having to deal with the possibility of scheme being empty.

In regards to the example in the issue, I think it is correct to interpret `//en.wikipedia.org/interesting_article` as an `origin-form` (as in the Actual Output). I just checked what Firefox sends when opening the URL `http://example.com//en.wikipedia.org/interesting_article`, it's

```
GET //en.wikipedia.org/interesting_article HTTP/1.1\r\n
```

Insomniak47:
I agree that this doesn't match the spec for request target. Though @seanmonstar has mentioned that isn't the be all and end all goal of the lib. I personally think it's an ergonomics improvement but I'm biased because I noticed this doing some web scraping and that was the motivation for the fix. If I'm reading RFC3986 S3.3 right it looks like path with no authority isn't allowed to start with two '/'s either. Which would mean it would have to return a parsing error or implement a different solution. I think the former case would be unintuitive.

For comparison: 
[...]

PR #301, opened 2019-03-09
Implementation of test and parsing for protocol relative uris (#300)

Insomniak47:
Implemented ability to handle protocol relative URIs (#300 )

You'll have to forgive me because I'm just starting out with rust so some of this might not be idiomatic and the test macro framework was/is a bit confusing.

Adding support for protocol relative uris required:
- The assumption that something that indicates scheme is always separated by three characters be turfed.
  - In places that this had an impact I've introduced locals with a name like offset or delimiter width and adjusted some of the magic 3's to use it instead. 
- Scheme matching on // for scheme2::parse and parse_exact
- Changing how as_str works to be able to output uris without a ":"

I ran benches locally and none of them seemed to be impacted by the additional branching

### DIFF
--- a/src/uri/mod.rs
+++ b/src/uri/mod.rs
@@ -924,22 +924,20 @@ impl PartialEq<str> for Uri {
         let mut absolute = false;
 
         if let Some(scheme) = self.scheme_part() {
-            let scheme = scheme.as_str().as_bytes();
-            absolute = true;
             
-            //Relative Protocol URI
-            let is_relative_protocol = scheme.len() == 0;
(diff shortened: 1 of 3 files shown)


slagroom:
good call; this is useful and should be considered

carllerche:
Thanks. This looks good to me. I would consider adding a variant to `Scheme2` to represent this case instead of checking for `len == 0`. Thoughts?

Insomniak47:
I think that makes sense. That was actually how I was approaching it initially but I got a bit thrown around by the design. I'll re implement with Relative as a Scheme2

PR #305, opened 2019-03-15
Create link references to `HeaderName` in `HeaderMap`

lzutao:
Note that this would only display `http::header::HeaderMap` when searching.
image

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -179,6 +179,7 @@ mod extensions;
 pub use convert::HttpTryFrom;
 pub use error::{Error, Result};
 pub use extensions::Extensions;
+#[doc(no_inline)]
 pub use header::HeaderMap;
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -179,6 +179,7 @@ mod extensions;
 pub use convert::HttpTryFrom;
 pub use error::{Error, Result};
 pub use extensions::Extensions;
+#[doc(no_inline)]
```
seanmonstar: Why preventing the inline here? I mostly just don't know what this does...

lzutao: More about it https://doc.rust-lang.org/rustdoc/the-doc-attribute.html#docno_inlinedocinline

seanmonstar:
Thank you!

PR #307, opened 2019-03-27
Impl Display on Version, impl Debug for Status

Gymmasssorla:
These changes will help to format the existing code abstractions into strings. I have derived Debug for Status and Version, and implement Display for Version. Hope this will help you.


src/status.rs
```diff
@@ -184,12 +183,6 @@ impl StatusCode {
     }
 }
 
-impl fmt::Debug for StatusCode {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        fmt::Debug::fmt(&self.0, f)
```
seanmonstar: This impl was preferred because seeing `status: 200` is cleaner than `status: StatusCode(200)`.

src/version.rs
```diff
@@ -54,15 +65,8 @@ impl Default for Version {
...
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        use self::Http::*;
-
-        f.write_str(match self.0 {
-            Http09 => "HTTP/0.9",
-            Http10 => "HTTP/1.0",
-            Http11 => "HTTP/1.1",
-            H2     => "HTTP/2.0",
-        })
+impl Display for Version {
```
seanmonstar: We specifically *didn't* implement `Display` because there isn't a canonical way to serialize a version. For instance, the HTTP2 identifier is "h2", but it's wrong to write `GET / h2`.

dadleyy: 👋 @seanmonstar! quick question - is this true for the `Response` type itself? That `Display` was intentionally _not_ implemented due to questions around what it would mean to implement for HTTP/2.0 responses?

I did a [quick impl](https://github.com/hyperium/http/compare/master...dadleyy:master) that formats the response for HTTP/1.0 correctly I think

seanmonstar: Yes, it's true for `Request` and `Response` as well.

PR #316, opened 2019-05-27
Add extensions_mut to request/response builders

sagebind:
In order to mutate existing extensions already present on a builder, add an `extensions_mut()` method that allows you to access previously set extensions.

My use-case is to create extension traits for `request::Builder` which adds methods that mutate special extensions added on the request. But in order to do that, I must be able to access my extension multiple times before the request is built.


seanmonstar:
Sorry to have missed this, this was done in #403.

PR #320, opened 2019-06-10
Remove/rename deprecated APIs

seanmonstar:


src/error.rs
```diff
@@ -84,7 +84,7 @@ impl error::Error for Error {
     // Return any available cause from the inner error. Note the inner error is
     // not itself the cause.
     #[allow(deprecated)]
-    fn cause(&self) -> Option<&error::Error> {
+    fn cause(&self) -> Option<&dyn error::Error> {
```
sfackler: Move to `source`?

PR #324, opened 2019-07-01
Refactored header name matching for WASM target

matiu2:
Advantages:

 1. Shorter code
 2. More rusty and less macroy
 3. It should work in wasm now untested and make yew users happier
    https://github.com/DenisKolodin/yew/issues/478

(Un?)fortunately I also ran rustfmt on it - which changed the layouts for the tables a bit but not hideously.

### DIFF
--- a/benches/header_name.rs
+++ b/benches/header_name.rs
@@ -1,0 +1,157 @@
+#![feature(test)]
+
+extern crate bytes;
+extern crate http;
+extern crate test;
+
+use http::header::HeaderName;
+use test::Bencher;
+
+fn make_all_known_headers() -> Vec<Vec<u8>> {
+    // Standard request headers
+    vec![b"A-IM".to_vec(),
+    b"Accept".to_vec(),
+    b"Accept-Charset".to_vec(),
+    b"Accept-Datetime".to_vec(),
+    b"Accept-Encoding".to_vec(),
+    b"Accept-Language".to_vec(),
+    b"Access-Control-Request-Method".to_vec(),
+    b"Authorization".to_vec(),
+    b"Cache-Control".to_vec(),
+    b"Connection".to_vec(),
+    b"Permanent".to_vec(),
+    b"Content-Length".to_vec(),
+    b"Content-MD5".to_vec(),
+    b"Content-Type".to_vec(),
+    b"Cookie".to_vec(),
+    b"Date".to_vec(),
+    b"Expect".to_vec(),
+    b"Forwarded".to_vec(),
+    b"From".to_vec(),
+    b"Host".to_vec(),
+    b"Permanent".to_vec(),
+    b"HTTP2-Settings".to_vec(),
+    b"If-Match".to_vec(),
+    b"If-Modified-Since".to_vec(),
+    b"If-None-Match".to_vec(),
+    b"If-Range".to_vec(),
+    b"If-Unmodified-Since".to_vec(),
+    b"Max-Forwards".to_vec(),
+    b"Origin".to_vec(),
+    b"Pragma".to_vec(),
+    b"Proxy-Authorization".to_vec(),
+    b"Range".to_vec(),
+    b"Referer".to_vec(),
+    b"TE".to_vec(),
+    b"User-Agent".to_vec(),
+    b"Upgrade".to_vec(),
+    b"Via".to_vec(),
+    b"Warning".to_vec(),
+    // common_non_standard
+    b"Upgrade-Insecure-Requests".to_vec(),
+    b"Upgrade-Insecure-Requests".to_vec(),
+    b"X-Requested-With".to_vec(),
+    b"DNT".to_vec(),
+    b"X-Forwarded-For".to_vec(),
+    b"X-Forwarded-Host".to_vec(),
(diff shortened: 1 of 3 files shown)


carllerche:
Hi, thanks for the work.

Would it be possible to include benchmarks for parsing header names? Probably could add a bench parsing each one + some common unknown header names. It would be interesting to compare the before and after.

matiu2:
I tried to make it work in nightly and beta, but then it broke for rust 1.20.

The existing benches are pretty much the same, for example:

New: `test basic::insert_500_custom_headers::header_map            ... bench:      35,208 ns/iter (+/- 3,372)`

Old: `test basic::insert_500_custom_headers::header_map            ... bench:      36,283 ns/iter (+/- 573)`

I'll some benches just for the name parsing..

matiu2:
I don't know how to bench unless it's running rust nightly (cargo bench) - so I had to make it work in nightly to get the numbers. It seems pretty close:

`cargo bench header_name`

# old macro style
test header_name_bad     ... bench:          15 ns/iter (+/- 0)
test header_name_easy    ... bench:          16 ns/iter (+/- 2)
test header_name_various ... bench:      10,222 ns/iter (+/- 991)

# new giant match statement
test header_name_bad     ... bench:          20 ns/iter (+/- 1)
test header_name_easy    ... bench:          21 ns/iter (+/- 1)
test header_name_various ... bench:       9,909 ns/iter (+/- 1,133)

benches/header_name.rs
```diff
@@ -0,0 +1,158 @@
...
+    b.iter(|| {
+        HeaderName::from_bytes(&name[..]).expect_err("Bad header name");
+    });
+}
+
+#[bench]
+fn header_name_various(b: &mut Bencher) {
+
+    b.iter(|| {
+        let all_known_headers = make_all_known_headers();
```
seanmonstar: This could be lifted out of the `iter`, so it isn't part of the measurements.

matiu2:
I've moved the make_all_known_headers out of the iterator. Sorry for not noticing that.

This commit is definitely slowing the code down by about 6% on my laptop :(

# New bench mark results

## Old code:

test header_name_bad     ... bench:          15 ns/iter (+/- 11)
test header_name_easy    ... bench:          16 ns/iter (+/- 1)
test header_name_various ... bench:       3,689 ns/iter (+/- 290)

## New code:

test header_name_bad     ... bench:          20 ns/iter (+/- 0)
test header_name_easy    ... bench:          24 ns/iter (+/- 1)
test header_name_various ... bench:       3,921 ns/iter (+/- 395)

--------

The main reason for this commit is so we can use `http` in wasm without an optimized build.

When debug building, those macros that expand out to `...&& b[0 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 + 1] == b'p'...` get put as-is into wasm, causing the wasm implementiation to blow up at run-time trying to add all the ones.

Here's the issue in `yew` that I'm trying to fix: https://github.com/DenisKolodin/yew/issues/478#issuecomment-507429151

matiu2:
Still slow. I don't know why :(

# Latest bench mark results

## Old macro based code:

test header_name_bad     ... bench:          15 ns/iter (+/- 0)
test header_name_easy    ... bench:          16 ns/iter (+/- 0)
test header_name_various ... bench:       3,648 ns/iter (+/- 123)

## New giant match statement code:

test header_name_bad     ... bench:          25 ns/iter (+/- 0)
test header_name_easy    ... bench:          24 ns/iter (+/- 0)
test header_name_various ... bench:       4,568 ns/iter (+/- 153)

matiu2:
Telling it the len of the array makes it a bit faster:

test header_name_bad     ... bench:          25 ns/iter (+/- 0)
test header_name_easy    ... bench:          24 ns/iter (+/- 0)
test header_name_various ... bench:       4,566 ns/iter (+/- 206)

matiu2:
# Latest bench mark results

## Old macro based code:

    $ cargo bench header_name
    ...
    test header_name_bad     ... bench:          14 ns/iter (+/- 0)
    test header_name_easy    ... bench:          16 ns/iter (+/- 1)
    test header_name_various ... bench:       3,763 ns/iter (+/- 80)

## New giant match statement code:

    test header_name_bad     ... bench:          25 ns/iter (+/- 1)
    test header_name_easy    ... bench:          24 ns/iter (+/- 0)
    test header_name_various ... bench:       4,227 ns/iter (+/- 75)

carllerche:
We should definitely get this working with wasm. Perhaps there is a way to conditional compile a different fn in wasm?

matiu2:
This version uses conditional compilation. It'll only compile the wasm version if there are 'debug_assertions' and the target is 'wasm32'.

It make the code longer, and harder to maintain though, because now if you add a new header, you'll have to add it in two places.

I think simplifying the code to the giant match statement is best. Although 8 more nanoseconds per header check may seem a lot; in msecs it is 8e-6 (0.0000008 msecs). But I'll leave the decision in your capable hands.

I tested it and it works in my wasm app in debug mode :)

PR #326, opened 2019-07-02
Implement HttpTryFrom<HashMap<String,String>> for HeaderMap

### LINKED ISSUE #325: Implement an easy conversion from HashMap to HeaderMap
I would be nice to be able to easily convert from a `HashMap<String, String>` to a `HeaderMap`. This might be implemented in a generic way with for instance an 

```rust
impl<COLLECTION, K, V> HttpTryFrom<COLLECTION> for HeaderMap<HeaderValue>
    where
        COLLECTION: IntoIterator<Item=(K,V)>,
        K: AsRef<str>,
        V: AsRef<str>
```

See : https://github.com/seanmonstar/reqwest/issues/555

What do you think ? Would you accept a pull request implementing that ?

lovasoa:
This allows an easy conversion from HashMap<String,String>
to HeaderMap<HeaderValue>.

Doing this conversion used to be unnecessarily verbose
because of the required mapping and error type conversions.

The implementation is generic:

```rust
    impl<COLLECTION, K, V> HttpTryFrom<COLLECTION> for HeaderMap<HeaderValue>
    where
        COLLECTION: IntoIterator<Item=(K, V)>,
        K: AsRef<str>,
        V: AsRef<str>
```

so it also works for HashMap<HeaderName, String> and for vectors.

Fixes #325

### DIFF
--- a/src/header/map.rs
+++ b/src/header/map.rs
@@ -1,14 +1,15 @@
-use super::HeaderValue;
-use super::name::{HeaderName, HdrName, InvalidHeaderName};
-use convert::{HttpTryFrom, HttpTryInto};
-use Error;
-use sealed::Sealed;
-
 use std::{fmt, mem, ops, ptr, vec};
+use std::collections::{BTreeMap, HashMap};
 use std::collections::hash_map::RandomState;
-use std::hash::{BuildHasher, Hasher, Hash};
+use std::hash::{BuildHasher, Hash, Hasher};
 use std::iter::FromIterator;
 use std::marker::PhantomData;
+
+use convert::{HttpTryFrom, HttpTryInto};
+use Error;
+
+use super::HeaderValue;
+use super::name::{HdrName, HeaderName, InvalidHeaderName};
 
 pub use self::as_header_name::AsHeaderName;
 pub use self::into_header_name::IntoHeaderName;
@@ -1724,7 +1725,23 @@ impl<T> FromIterator<(HeaderName, T)> for HeaderMap<T>
     }
 }
 
-impl<T> Sealed for HeaderMap<T> {}
+/// A marker trait that allows conversion from a type to HeaderMap
+/// This trait is needed to have both a fast implementation of
+/// HttpTryFrom<HeaderMap> and a generic implementation
+/// of HttpTryFrom<IntoIterator<Item=(K,V)>> for HeaderMap that do not conflict.
+pub trait IntoHeaderMapAllowed {}
+
+impl<K, V, S> IntoHeaderMapAllowed for HashMap<K, V, S> {}
+
+impl<'a, K, V, S> IntoHeaderMapAllowed for &'a HashMap<K, V, S> {}
+
+impl<K, V> IntoHeaderMapAllowed for BTreeMap<K, V> {}
+
+impl<'a, K, V> IntoHeaderMapAllowed for &'a BTreeMap<K, V> {}
+
+impl<'a, T> IntoHeaderMapAllowed for &'a [T] {}
+
+impl<T> IntoHeaderMapAllowed for Vec<T> {}
 
 /// Convert a collection of tuples into a HeaderMap
 ///
@@ -1746,7 +1763,7 @@ impl<T> Sealed for HeaderMap<T> {}
 /// ```
 impl<C, K, V> HttpTryFrom<C> for HeaderMap<HeaderValue>
     where
-        C: IntoIterator<Item=(K, V)>,
+        C: IntoIterator<Item=(K, V)> + IntoHeaderMapAllowed,
         HeaderName: HttpTryFrom<K>,
         HeaderValue: HttpTryFrom<V>
 {
--- a/src/header/map.rs
+++ b/src/header/map.rs
@@ -1745,6 +1745,7 @@ impl<T> FromIterator<(HeaderName, T)> for HeaderMap<T>
 /// ```
 impl<'a, K, V> HttpTryFrom<&'a HashMap<K, V>> for HeaderMap<HeaderValue>
     where
+        K: Eq + Hash,
         HeaderName: HttpTryFrom<&'a K>,
         HeaderValue: HttpTryFrom<&'a V>
 {
--- a/src/header/map.rs
+++ b/src/header/map.rs
@@ -1,6 +1,6 @@
 use super::HeaderValue;
 use super::name::{HeaderName, HdrName, InvalidHeaderName};
-use convert::HttpTryFrom;
+use convert::{HttpTryFrom, HttpTryInto};
 use Error;
 use sealed::Sealed;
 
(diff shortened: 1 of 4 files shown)


lovasoa:
The broken CI is not due to the changes introduced in this PR, but to #327.

seanmonstar:
Hm, what if instead of K and V being `AsRef`, they were `HeaderName: HttpTryFrom<K>` and `HeaderValue: HttpTryFrom<V>`?

carllerche:
:+1: on the idea.

lovasoa:
@seanmonstar OK, but I will have to add some `HttpTryFrom` implementations to HeaderName and HeaderValue, then, if we want it to still work with `HashMap<String,String>`.

lovasoa:
@seanmonstar : I did the change you suggested. It required adding an impl HttpTryFrom<&String> for HeaderName and HeaderValue.

lovasoa:
@seanmonstar Is it ok like that, or is there still something you would like me to change ?

lovasoa:
Any news?

src/header/map.rs
```diff
@@ -23,10 +26,11 @@ pub use self::into_header_name::IntoHeaderName;
 /// ```
 /// # use http::HeaderMap;
 /// # use http::header::{CONTENT_LENGTH, HOST, LOCATION};
-/// let mut headers = HeaderMap::new();
-///
-/// headers.insert(HOST, "example.com".parse().unwrap());
-/// headers.insert(CONTENT_LENGTH, "123".parse().unwrap());
+/// # use http::HttpTryFrom;
```
seanmonstar: I'd leave the example the way it is, as it shows the normal `insert` method.

src/header/map.rs
```diff
@@ -1720,6 +1724,45 @@ impl<T> FromIterator<(HeaderName, T)> for HeaderMap<T>
...
+/// ].iter().cloned().collect();
+///
+/// let good_headers: Result<HeaderMap> = HeaderMap::try_from(&headers_hashmap);
+/// assert!(good_headers.is_ok());
+///
+/// headers_hashmap.insert("\r".into(), "\0".into());
+/// let bad_headers: Result<HeaderMap> = HeaderMap::try_from(&headers_hashmap);
+/// assert!(bad_headers.is_err());
+/// ```
+impl<COLLECTION, K, V> HttpTryFrom<COLLECTION> for HeaderMap<HeaderValue>
```
seanmonstar: Small nit, not a blocker, but my personal taste is to have `COLLECTION` just be `C`.

src/header/map.rs
```diff
@@ -1720,6 +1724,45 @@ impl<T> FromIterator<(HeaderName, T)> for HeaderMap<T>
     }
 }
 
+impl<T> Sealed for HeaderMap<T> {}
```
seanmonstar: There should probably be an identity impl, `impl<T> HttpTryFrom<HeaderMap<T>> for HeaderMap<T>`. Would that conflict with the generic one below?

lovasoa: Yes, `HeaderMap<T>` is itself `IntoIterator<Item=(HttpTryFrom<HeaderName>, HttpTryFrom<HeaderValue>)>`, so `HeaderMap::try_from(header_map)` works as expected.

lovasoa:
@seanmonstar I implemented your the changes you suggested. Hope this can be merged soon.

src/header/map.rs
seanmonstar: Usually one would expect `TryFrom<Me> for Me` to be basically free, but because that conflicts, it would be surprising that it essentially clones a new `HeaderMap`....

seanmonstar:
It's worth noting that the current conversion can be done fairly ergonomically already:

```rust
let mut map = HashMap::new();
map.insert("hello", "world");
map.insert("and", "good day");
    
let headers = map
    .into_iter()
    .map(|(name, value)| {
        Ok((
            name.parse()?,
            value.parse()?,
        ))
    })
    .collect::<Result<HeaderMap, http::Error>>();
```

lovasoa:
@seanmonstar The main advantage of implementing HttpTryFrom is not only to remove all these lines of boilerplate, but also to allow libraries to provide functions that accept a trait instead of having the users manually converting to custom types all the time. Also, shouldn't it have been discussed earlier ?

seanmonstar:
> shouldn't it have been discussed earlier ?

True, I'm very sorry I didn't realize how to make the original request shorter. I didn't think there would be a conflict with a generic `HttpTryFrom` implementation.

I recognize what the goal is, I'm just pointing out what exists already. I believe that the identity implementation isn't "free" would be surprising to everyone.

src/header/map.rs
lovasoa: This is indeed sub-optimal, and can fixed later when rust adds support for [specialization](https://github.com/rust-lang/rust/issues/37653).
But until then, I think the usability benefits far outweigh this downside. The standard library itself has the same issue in several places, where an implementation is slower than it could be in order to be more generic.

lovasoa:
@seanmonstar I answered [above](https://github.com/hyperium/http/pull/326#discussion_r302779661). I see what you mean, but I don't think this should block merging this. Do you ?

lovasoa:
I prefer the generic option, but if you think the slight HeaderMap to HeaderMap conversion overhead is a real problem people would encounter, we can switch to having a separate implementation for a few standard library containers (maybe only Vec and HashMap), and that wouldn't conflict with `HttpTryFrom<HeaderMap> for HeaderMap`.

seanmonstar:
A reason I think it'd be surprising is, for example, if some `Builder` were to change their `headers` method to one of `HeaderMap: HttpTryFrom<T>`, a user who already had a `HeaderMap` would end up paying twice.

For the completely generic situation, I think the code I pasted a few comments ago is ok. If wanting to add `HttpTryFrom` anyways, it's probably best to just provide it for concrete `std` collections for now, like `HashMap`. Providing one for `Vec<(K, V)>` seems fairly specific, is it that common?

lovasoa:
`Vec<(K,V)>` would allow using the standard `vec!` macro, which allows readable one-line definitions of headers. I can omit it if you want, but I see no reason to.

lovasoa:
@seanmonstar I added a marker trait to limit the generic implementation to just a few standard library types, so that the generic implementation does not apply to `HeaderMap` itself. Tell me what you think ?

seanmonstar:
I appreciate the goal, but I think there's value in staying conservative, because we can't really make breaking changes, since `http` is meant to be an interoperable crate for the whole ecosystem.

We could start with just `HashMap`, and then adding others in the future is not a breaking change. How about we start there?

lovasoa:
@seamonstar I removed the implementation for non-HashMap types. I hope this can be merged now.

seanmonstar:
@lovasoa Thank you! I see this is targeting master, but I assume you want to use this before 0.2 releases? If so, if you can change the target to `hyperium:0.1.x`, I can merge to the right branch.

Would you be able to remove the marker trait for now, since we'll only have the `HashMap` implementation to start, so as to not have more in the public API? Or I can do so as well during the merge.

lovasoa:
Ok, I'll change the target of the PR. I think the marker trait is still needed to implement the conversion from both HashMap and &HashMap. Or is there another way?

lovasoa:
image

There is no `0.1.x` branch...

seanmonstar:
Woops, I'm so sorry, I got confused. Most of the other repos I'm working in have converted master to the newer version, but this one still has master as 0.1. I'll merge this in first, and then update to be more consistent.

seanmonstar:
Er, it seems there was no `into_iter` for `&HashMap` in Rust 1.20?

lovasoa:
@seanmonstar I fixed rust 1.20 compatibility. I think this is ready for merging.

PR #328, opened 2019-07-02
Improve CI

### LINKED ISSUE #327: Compilation is broken on nightly
Compilation using the latest nightly (1.37.0-nightly (17e62f77f 2019-07-01)) emits warnings, and this prevents compilation because of `#![deny(warnings)]`.

lovasoa:
- Allow warnings on rust nightly (this allows keeping compatibility with both rust 1.20 and nightly)
 - Create and independent script for CI instead of writing code inside .travis.yml
 - Fix warnings in the tests

Fixes #327

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,18 +8,15 @@ matrix:
     - rust: stable
     - rust: beta
     - rust: nightly
+      env: DO_BENCHMARKS=1 ALLOW_WARNINGS=1
     # ensure wasm always builds
     - rust: stable
-      script:
-        - rustup target add wasm32-unknown-unknown
-        - cargo build --target=wasm32-unknown-unknown
+      env: TARGET=wasm32-unknown-unknown SKIP_TESTS=1
     # minimum rustc version
     - rust: 1.20.0
-      script: cargo build
+      env: SKIP_TESTS=1
 
-script:
-  - cargo test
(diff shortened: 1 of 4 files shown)


seanmonstar:
I think this means that when that lint makes it to beta (in a couple days?), we'll see it triggered there as well.

How about instead of removing the `deny(warnings)` entirely, just adding an allow where a warning we can't fix it triggered. Like:

```rust
#[allow(warnings)]
type AnyMap = HashMap<TypeId, Box<Any + Send + Sync>, BuildHasherDefault<IdHasher>>;
```

lovasoa:
> How about instead of removing the deny(warnings) entirely, just adding an allow where a warning we can't fix it triggered.

`#[allow(warnings)]` would allow any kind of warning on the portion of code it is applied to, not just compatibility warnings. Is that really what we want ?

> I think this means that when that lint makes it to beta (in a couple days?), we'll see it triggered there as well.

Maybe warnings should be forbidden only on stable, then ?

lovasoa:
on a side note: another change that this PR introduces is that warnings are disallowed even in tests.

seanmonstar:
> would allow any kind of warning on the portion of code it is applied to, not just compatibility warnings. Is that really what we want ?

Yep, that's fine. It's a smaller change, and there's no concern of other warnings being triggered.

lovasoa:
Feel free to close or edit this PR if you think it is not relevant, then.

PR #331, opened 2019-07-17
Upgrade to 2018 Edition

davidbarsky:
Mostly automated change. I left out some of the `rustfmt` changes, as they impacted some hand-written byte tables.


davidbarsky:
@seanmonstar Would be happy to! However, I thought that `http` allows for a squash on merge like `tracing` does?

seanmonstar:
Oh we could squash it all into one if you'd rather!

davidbarsky:
Oh, thanks for taking care of that! I assumed `http` prefers a squash-and-merge strategy.

PR #332, opened 2019-07-23
Change request, response, and uri Builders to be by-value

### LINKED ISSUE #191: Move Builders to by-value
I propose that `http::request::Builder` and `http::response::Builder` should be changed to be by-value builders, instead of by-ref as they currently are. This may be a little controversial, but I believe there's good reason to do so, and my proposal still includes ways to mutate builders with `&mut self` for cases that want it.

## Motivation

- **Ergonomics**:
  - Using by-ref builders is easy when you can write all method calls in a single chain, but as soon as the chain must be broken, it must be broken into *at least* 3 pieces: `let mut b = Request::builder()`, `b.chain().chain();`, and then `b.chainMore().build()`.
  - Creating a builder, and setting initial properties, before passing it as a return value or to a function *requires* you to break up the chain as mentioned in the previous point, since you must get the original value from `builder()` to be able to pass it anywhere by value.
- **Correctness**:
  - Using by-ref builders means that calling `build` leaves behind a mine in the builder slot, since the builder must consume internal state to build. That means the compiler will not catch any attempts to use the builder *after* a call to `build`.
  - Passing `&mut Builder` to a function can actually consume the internal builder state, which can result in unexpected panics after having passed it to a modifier where the assumption was it was just set some properties.

### Improvements

- **Ergonomics**:
  - With a by-value builder, it is easy to stop the chain at any point and pass the `Builder` by value to another function or return it. It is not required to break up the chain to appease the type system.
- **Correctness**:
  - If a user has a `Builder`, they know for certain that it is not a consumed shell that might panic if used more. Using the builder to `build` requires ownership of it, so supposed modifier functions cannot accidentally consume it.

### But conditionally chaining on by-value is annoying

First, I'd offer that if we have to pick between two annoying things, dealing with a by-value builder is the *less* annoying of the two. Compare:

```rust
// by-ref
let mut builder = Request::builder();
builder
    .uri("/foo/bar")
    .method("POST");

if body_len != 0 {
    builder.header("content-length", body.to_string());
}

builder
    .body(body)
    .unwrap()
```

VS

```rust
// by-value
let mut builder = Request::builder()
    .uri("/foo/bar")
    .method("POST");

if body_len != 0 {
    builder = builder.header("content-length", body.to_string());
}

builder
    .body(body)
    .unwrap()
```

However, as I said in the beginning, we can add methods that mutate on `&mut self`, to allow that pattern for those who want it. Basically, every builder method that mutates self by-value, like `uri`, we could add `set_uri(&mut self, ...) -> &mut Self` as a mirror.

These by-ref mutators could be chained as well. The only pain is that they couldn't be chained into `body()`, since that would require by-value. However, I've provided above while I feel this is better.

### Alternative

We could optionally have the by-ref mutates keep the same name, and make the by-value methods have a `with_` prefix. However, this would end up as a bigger breaking change, since **all** existing code could no longer call `body()` on their chains.

## Breaking Changes

If we switch to by-value without the prefix, the breaking changes would be these:

- Using methods when receiving a `&mut Builder` would need to be changed to the `set_` versions.
- Anyone (ab)using being able to construct a the item from a `&mut Builder` would no longer be able to do so.

However, anyone that has the full chain together should **not** see any breaking change.

seanmonstar:
Since the builders are only 1-time use anyways, it makes more sense to
enforce that through the type system, than to leave behind a panic bomb.

Most code that was just one chain should be un-affected.

Closes #191

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -12,12 +12,12 @@
 //! ```no_run
 //! use http::{Request, Response};
 //!
-//! let mut request = Request::builder();
-//! request.uri("https://www.rust-lang.org/")
-//!        .header("User-Agent", "my-awesome-agent/1.0");
+//! let mut request = Request::builder()
+//!     .uri("https://www.rust-lang.org/")
+//!     .header("User-Agent", "my-awesome-agent/1.0");
 //!
 //! if needs_awesome_header() {
-//!     request.header("Awesome", "yes");
+//!     request = request.header("Awesome", "yes");
 //! }
 //!
 //! let response = send(request.body(()).unwrap());
@@ -58,7 +58,7 @@ use std::fmt;
 use crate::header::{HeaderMap, HeaderName, HeaderValue};
 use crate::method::Method;
 use crate::version::Version;
-use crate::{Error, Extensions, HttpTryFrom, Result, Uri};
+use crate::{Extensions, HttpTryFrom, Result, Uri};
 
 /// Represents an HTTP request.
 ///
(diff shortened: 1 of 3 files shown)


src/request.rs
```diff
@@ -938,35 +895,26 @@ impl Builder {
...
-            None => None,
-        }
+        self.inner.as_ref().ok().map(|h| &h.headers)
     }
 
-    /// Get header on this request builder.
-    /// when builder has error returns None
+    /// Get headers on this request builder.
+    ///
+    /// When builder has error returns None.
```
LucioFranco:
```suggestion
    /// Returns None when builder as an error.
```

PR #333, opened 2019-07-24
[0.2.x] introduce TryFrom impls for path, scheme, uri and authority

prasannavl:
This is a redo of https://github.com/hyperium/http/pull/308 on the 0.2.x branch

- This PR introduces `TryFrom`, however doesn't remove `HttpTryFrom` or change other bits in any way, to keep the PR small and unblock others to continue this work. 
- Further PRs can remove `HttpTryFrom` completely if so required. 

/cc @seanmonstar @davidbarsky


davidbarsky:
Thanks for this! I’m not one with the final authority on the matter, but I’d be happy to merge this in and pick up the remaining work of removing `HttpTryFrom`.

Issue #144, opened 2017-11-17
Support custom status code


kamyuentse:
Currently, hyper and this crate only support status code from `[100, 600)`, otherwise will cause error `InvalidStatusCode`, which make parse a `Response` with custom code failed eventually.

For example:
```
HTTP/1.1 6xx status code 6xx

....
```
will cause
```
TRACE:hyper::proto::conn                        : Conn::read_head
TRACE:hyper::proto::h1::parse                   : Response.parse([Header; 100], [u8; 447])
TRACE:hyper::proto::h1::parse                   : Response.parse Complete(409)
TRACE:hyper::proto::conn                        : State::close_read()
DEBUG:hyper::proto::conn                        : parse error (Invalid Status provided) with 447 bytes
TRACE:tokio_proto::streaming::pipeline::advanced: process_out_frame
ERROR:tokio_proto::streaming::pipeline::client  : pipeline error: An error occurred.
```

But, sometime, some service will define custom status code, especially in some RESTful API.

So, I think it is fine to support custom code and reason.

carllerche:
A status code of 6xx violates the HTTP spec.

Do you have an example of another server returning 6xx?

kamyuentse:
I know that neither HTTP spec nor IANA defines the status code greater than 599.

As I mentioned above, some RESTful service may use UNREGISTERED status codes.

The following is an example which uses `openresty` as backend server:
```
curl -v http://uc.qbox.me/v2/query\?ak\=INVALID_KEY\&bucket\=INVALID_BUCKET
```

Not support custom status codes and response phrase will stumble those want to migrate to rust.

steveklabnik:
https://tools.ietf.org/html/rfc7231#section-8.2.2

> New status codes are required to fall under one of the categories defined
in Section 6 <https://tools.ietf.org/html/rfc7231#section-6>.

kamyuentse:
* What you mean is that we MUST follow the HTTP spec;
* But what I am talking about is providing the ability to make user can define custom status code inside their business, as long as the participant in the communication know what they mean;

I think there are no conflict.

curl, openresty, wireshark ... and so on, can parse the undefined status code and response phrase, why `http` and `hyper` have a limitation here.

carllerche:
By sticking with the spec, the http crate is able to make performance optimizations. Violating the spec in this way would make the crate slower for those who follown then spec. 

The argument to make the library slower for everyone would need to be very compelling.

quininer:
[...]

PR #336, opened 2019-08-13
Preserve invalid status code at best effort

upsuper:
This should help on usecases mentioned in #144 without adding performance penalty for everyone else.

Theoretically with the original form, we can use `NonZeroU16` and Rust would optimize `Result<StatusCode, InvalidStatusCode>` to just a `u16`, which is probably unbeatable. But it isn't clear whether returning `Result<u16, ()>` could really be faster than `Result<u16, u16>`.

If that optimization does matter, we can probably hide the new field behind a feature and have people want it to opt in.

### DIFF
--- a/src/status.rs
+++ b/src/status.rs
@@ -46,10 +46,11 @@ pub struct StatusCode(u16);
 ///
 /// This error indicates that the supplied input was not a valid number, was less
 /// than 100, or was greater than 599.
+///
+/// Invalid value converted from `u16` is preserved. Invalid value converted from `&str` is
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for the PR! I see now by the implementation you mean to preserve the `u16` which is out-of-range in the `InvalidStatusCode` error type. Do you feel this provides substantial benefits over being able to just look at the original `u16` before calling `StatusCode::from_u16`?

upsuper:
After some discussion, I don't think this is going to work for the usecase. Closing now.

Sorry for the noise.

PR #337, opened 2019-08-14
Switch remaining HttpTryFrom to TryFrom

davidbarsky:
I figured I should post this PR, but its failing to build due to type inference errors on the error in `src/request.rs` and `src/uri/builder.rs`. For instance:

```
   --> src/request.rs:785:26
    |
785 |             let method = TryFrom::try_from(method).map_err(Into::into)?;
    |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type

error[E0277]: `?` couldn't convert the error to `error::Error`
  --> src/uri/builder.rs:50:50
   |
50 |             parts.scheme = Some(scheme.try_into()?);
   |                                                  ^ the trait `std::convert::From<<uri::scheme::Scheme as std::convert::TryFrom<T>>::Error>` is not implemented for `error::Error`
   |
   = note: the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait
   = help: the following implementations were found:
             <error::Error as std::convert::From<header::name::InvalidHeaderName>>
             <error::Error as std::convert::From<header::name::InvalidHeaderNameBytes>>
             <error::Error as std::convert::From<header::value::InvalidHeaderValue>>
             <error::Error as std::convert::From<header::value::InvalidHeaderValueBytes>>
           and 5 others
   = note: required by `std::convert::From::from`
```

Any suggestions on resolving them?


src/header/name.rs
```diff
@@ -1759,32 +1759,23 @@ impl From<HeaderName> for Bytes {
     }
 }
 
-impl<'a> HttpTryFrom<&'a HeaderName> for HeaderName {
```
seanmonstar: Aren't the impls of `TryFrom<&'_ Self>` still needed, so as to clone itself?

davidbarsky: Rustc complains:

```
error[E0119]: conflicting implementations of trait `std::convert::TryFrom<&header::name::HeaderName>` for type `header::name::HeaderName`:
    --> src/header/name.rs:1778:1
     |
1778 | impl<'a> TryFrom<&'a HeaderName> for HeaderName {
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: conflicting implementation in crate `core`:
             - impl<T, U> std::convert::TryFrom<U> for T
               where U: std::convert::Into<T>;
```

Need-To-Learn:
@davidbarsky Unfortunately I got this errors too when i tried to migrate from HttpTryFrom, i didnt' find a solution yet.

davidbarsky:
Thanks to @seanmonstar's I've gotten this to compile.

src/error.rs
```diff
@@ -160,33 +163,11 @@ impl From<header::InvalidHeaderValueBytes> for Error {
...
-impl fmt::Display for Never {
-    fn fmt(&self, _f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match *self {}
-    }
-}
-
-impl error::Error for Never {
-    fn description(&self) -> &str {
-        match *self {}
+impl From<std::convert::Infallible> for Error {
```
davidbarsky: Sean correctly pointed out that this might be a forwards-compatibility hazard once `!` lands in Rust.

pyfisch: I'm curious how this may create forward-compatibility problems. [The docs](https://doc.rust-lang.org/std/convert/enum.Infallible.html#future-compatibility) only warn about implementing a trait for both Infallible *and* `!` but this does not appear to be the case here.

davidbarsky: Hmm, then I think this might be fine?

davidbarsky:
Hi @seanmonstar! I think this is ready to review.

src/error.rs
```diff
@@ -160,33 +163,11 @@ impl From<header::InvalidHeaderValueBytes> for Error {
...
+    fn from(err: std::convert::Infallible) -> Error {
+        Error {
+            inner: ErrorKind::Infallible(err),
```
seanmonstar: I don't believe this variant is needed, it should be sufficient to write `match err {}`.

seanmonstar:
Merged manually https://github.com/hyperium/http/commit/f264abddc4099b33a309d9ee84159103692db12a

PR #343, opened 2019-09-11
Allow % in IPv6 addresses to support zone identifiers.

### LINKED ISSUE #322: Zones in link-local IPv6 addresses seems to be not supported.
```rust
extern crate http;

fn main() {
    use http::uri::Uri;
    use std::str::FromStr;
    Uri::from_str("http://[fe80::1%25eth0]:8080/").unwrap();
}
```

`InvalidUri(InvalidAuthority)`.

It parses if I omit the zone index (`%25eth0`), but such URI is expected to be unconnectable by itself.

Should `http` crate support [RFC6874](https://tools.ietf.org/html/rfc6874) and allow link-local IPv6 in URLs?

ComputerDruid:
https://tools.ietf.org/html/rfc6874 explains that %25 can be used to
indicate a "zone identifier", useful for IPv6 link-local addresses.

Fixes #322

### DIFF
--- a/src/uri/authority.rs
+++ b/src/uri/authority.rs
@@ -627,4 +627,13 @@ mod tests {
         let result: Authority = authority_str.parse().unwrap();
         assert_eq!(result, authority_str);
     }
+
+    #[test]
+    fn rejects_percent_outside_ipv6_address() {
+        let err = Authority::parse_non_empty(b"1234%20[fe80::1:2:3:4]").unwrap_err();
+        assert_eq!(err.0, ErrorKind::InvalidAuthority);
+
+        let err = Authority::parse_non_empty(b"[fe80::1:2:3:4]%20").unwrap_err();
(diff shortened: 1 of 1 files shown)


ComputerDruid:
Note: https://github.com/hyperium/hyper/pull/1937 implements the support in hyper.

src/uri/authority.rs
```diff
@@ -103,15 +103,20 @@ impl Authority {
                 }
                 b':' => {
                     colon_cnt += 1;
-                },
+                }
                 b'[' => {
                     start_bracket = true;
+                    if has_percent {
+                        // Something other than the userinfo has a `%`, so reject it.
+                        return Err(ErrorKind::InvalidAuthority.into());
```
erickt: Can you add a test case that verifies this error gets returned?

ComputerDruid: Done.

PR #348, opened 2019-10-28
Refactor Method::is_idempotent

XAMPPRocky:
This refactors the method into a single condition to try to provide better code readability.


sfackler:
The changed implementation is broken. PUT and DELETE are idempotent but not safe.

PR #350, opened 2019-11-11
bytes: require 0.4.4

mathstuf:
Various APIs used in this crate were added in 0.4.4.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ keywords = ["http"]
 categories = ["web-programming"]
 
 [dependencies]
-bytes = "0.4"
+bytes = "0.4.13"
 fnv = "1.0.5"
 itoa = "0.4.1"
 


mathstuf:
Hrm. Seems `bytes` itself needs some work for minimal-versions in some cases. I'll have a followup within a few days.

mathstuf:
Requires a release of https://github.com/tokio-rs/bytes/pull/312

mathstuf:
I poked the bottom of this PR hole; hopefully it can cause some movement to let this get merged :) .

carllerche:
looks like a conflict ::(

mathstuf:
Not surprising; I'll rebase once the dependent PRs are merged (and have a release on crates.io to depend on).

dekellum:
Conflict with #364 and specifically 845fdd1?

mathstuf:
Well, this might need to go to 0.1.x as well now because of that :) .

dekellum:
Branch whack-a-mole!   Seriously, thanks for trying to fix this. I hope there is a potential for an 0.1.21 with the fix, as I suspect some projects will need to support both http 0.1 and 0.2 for some transition period.

mathstuf:
I'll probably be back for 0.2 at some point. I've got a lot of plates in the air for my own set of crates at the moment before I can start to *actually* fix everything :) .

mathstuf:
Ping?

LucioFranco:
@mathstuf is this something we still want to go with or should we close this in favor of ensuring 0.5 has all the fixes we need?

mathstuf:
Oh, my bad. I thought this was the bottom of my stack. This will need the PR mentioned above first. Sorry for the noise. Going down one level deeper.

kaj:
If there is action on #440 this should probably just be closed as semi-duplicate.

mathstuf:
I'm more interested in `minimum-versions` support, so if that fixes it, I'm fine with that.

robjtede:
this can be closed

mathstuf:
It looks like #440 also got closed. Would another minimal versions support PR make it any further?

seanmonstar:
Yea, it'd be a good thing to add to CI. Sorry we missed these, but we have more attention now.

Issue #354, opened 2019-11-16
Failing to drop `HeaderMap::Drain` causes double-free


Qwaz:
https://github.com/hyperium/http/blob/9c05e391e00474abaa8c14a86bcb0fc5eff1120e/src/header/map.rs#L2115-L2122

https://github.com/hyperium/http/blob/9c05e391e00474abaa8c14a86bcb0fc5eff1120e/src/header/map.rs#L2140-L2148

[Failing to drop a value is considered safe](https://doc.rust-lang.org/nomicon/leaking.html) in Rust, and unsafe code should not rely on this behavior.

> It is reasonable for safe code to assume that destructor leaks do not happen, as any program that leaks destructors is probably wrong. However unsafe code cannot rely on destructors to be run in order to be safe.

`HeaderMap::Drain` uses `ptr::read` to move out entries from the map when it iterates, and calls `map.entries.set_len(0)` to clear the map at once when it is dropped. If `Drain`'s drop is not called, double-free happens when `HeaderMap` is dropped. Also, if `Drain` is dropped without iterating to the end, it leaks memory.

[Demonstration](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c5d2488bd0ef3ea6826cc5b17d264845)

Shnatsel:
Double free is an [exploitable security vulnerability](https://heap-exploitation.dhavalkapil.com/attacks/double_free.html). After releasing a fixed version please file a [security advisory](https://github.com/RustSec/advisory-db) and, if possible, yank the affected versions from crates.io.

hawkw:
I started attempting to fix this using a similar approach as [`Vec::drain`](https://doc.rust-lang.org/nomicon/leaking.html#drain), setting the length of `self.entries` to 0 when creating the `Drain` iterator, and then setting the length back if the `Drain` iterator is dropped before draining all elements. This way, `mem::forget`ting the iterator would only amplify a leak, rather than calling a double-free. 

However, this isn't trivially possible here: the `ValueDrain` iterator calls `HeaderMap::remove_extra_values`, which relies on indexing the `entries` vector. If we've already set it's length to 0, the bounds checks there will fail. So, either that method needs to be refactored to not rely on indexing, or we need a different solution here (possibly involving `ManuallyDrop`?). I'll keep working on it.

PR #357, opened 2019-11-25
Fix HeaderMap::drain double-free if drain iterator is forgotten

seanmonstar:
Fixes https://github.com/hyperium/http/issues/354

cc @Qwaz thanks for the report!

### DIFF
--- a/src/header/map.rs
+++ b/src/header/map.rs
@@ -135,7 +135,9 @@ pub struct ValuesMut<'a, T: 'a> {
 #[derive(Debug)]
 pub struct Drain<'a, T: 'a> {
     idx: usize,
-    map: *mut HeaderMap<T>,
+    len: usize,
+    entries: *mut [Bucket<T>],
+    extra_values: *mut Vec<ExtraValue<T>>,
     lt: PhantomData<&'a mut HeaderMap<T>>,
 }
 
@@ -202,7 +204,8 @@ pub struct ValueIterMut<'a, T: 'a> {
 /// An drain iterator of all values associated with a single header name.
 #[derive(Debug)]
 pub struct ValueDrain<'a, T: 'a> {
-    map: *mut HeaderMap<T>,
+    raw_links: RawLinks<T>,
+    extra_values: *mut Vec<ExtraValue<T>>,
     first: Option<T>,
     next: Option<usize>,
     lt: PhantomData<&'a mut HeaderMap<T>>,
@@ -269,6 +272,13 @@ struct Links {
     next: usize,
     tail: usize,
 }
+
+/// Access to the `links` value in a slice of buckets.
+///
+/// It's important that no other field is accessed, since it may have been
+/// freed in a `Drain` iterator.
+#[derive(Debug)]
+struct RawLinks<T>(*mut [Bucket<T>]);
 
 /// Node in doubly-linked list of header value entries
 #[derive(Debug, Clone)]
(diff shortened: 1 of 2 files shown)


src/header/map.rs
```diff
@@ -1631,6 +1543,129 @@ impl<T> HeaderMap<T> {
...
+
+    #[inline]
+    fn raw_links(&mut self) -> RawLinks<T> {
+        RawLinks(&mut self.entries[..] as *mut _)
+    }
+}
+
+/// Removes the `ExtraValue` at the given index.
+#[inline]
+fn remove_extra_value<T>(mut raw_links: RawLinks<T>, extra_values: &mut Vec<ExtraValue<T>>, idx: usize) -> ExtraValue<T> {
```
hawkw: Nit/TIOLI: seems like this could be a method on `RawLinks`? Not a big deal regardless.

src/header/map.rs
```diff
@@ -270,6 +273,13 @@ struct Links {
     tail: usize,
 }
 
+/// Access to the `links` value in a slice of buckets.
+///
+/// It's important that no other field is accessed, since it may have been
+/// freed in a `Drain` iterator.
+#[derive(Debug)]
+struct RawLinks<T>(*mut [Bucket<T>]);
```
hawkw: Similarly, it looks like the pointer is never null (and is therefore never null checked); consider 
```suggestion
struct RawLinks<T>(ptr::NonNull<[Bucket<T>]>);
```
to encode that in the type system?

src/header/map.rs
```diff
@@ -135,7 +135,9 @@ pub struct ValuesMut<'a, T: 'a> {
 #[derive(Debug)]
 pub struct Drain<'a, T: 'a> {
     idx: usize,
-    map: *mut HeaderMap<T>,
+    len: usize,
+    entries: *mut [Bucket<T>],
+    extra_values: *mut Vec<ExtraValue<T>>,
```
hawkw: AFAICT, these are never null & are not null-checked; consider using `NonNull` to encode this in the type system.

src/header/map.rs
```diff
@@ -2958,7 +2997,10 @@ impl<'a, T> Iterator for ValueDrain<'a, T> {
             self.first.take()
         } else if let Some(next) = self.next {
             // Remove the extra value
-            let extra = unsafe { &mut (*self.map) }.remove_extra_value(next);
+            let extra = unsafe {
+                //&mut (*self.map) }.remove_extra_value(next);
```
hawkw: Nit: commented out code should probably be removed?

tests/header_map.rs
```diff
@@ -86,6 +86,40 @@ fn drain() {
...
+    headers.insert("zomg", "bar".parse().unwrap());
+    headers.append("hello", "world2".parse().unwrap());
+
+    let iter = headers.drain();
+    assert_eq!(iter.size_hint(), (2, Some(2)));
+    // not consuming `iter`
+}
+
+#[test]
+fn drain_forget() {
```
hawkw: Should there also be a test for `mem::forget`ting a `ValueDrain` returned by a `Drain`?

src/header/map.rs
seanmonstar: We could, though I'm not sure it gives much benefits here... It's mostly a layout optimization. but is otherwise more annoying to use.

hawkw: The motivation I had in mind was mostly "making it clearer to a future reader that these are never null", rather than optimizations. Up to you.

PR #360, opened 2019-11-25
fix capacity overflows in HeaderMap::reserve

### LINKED ISSUE #352: Size issue in `HeaderMap::reserve()`
https://github.com/hyperium/http/blob/9c05e391e00474abaa8c14a86bcb0fc5eff1120e/src/header/map.rs#L622-L640

[`usize::next_power_of_two()`](https://doc.rust-lang.org/std/primitive.usize.html#method.next_power_of_two) method silently overflows to 0 in release mode. This makes it possible to shrink the size of the map to 0 with `HeaderMap::reserve()`.

- If the map doesn't contain any entry, it sets the mask value to `usize::MAX` which is inconsistent but doesn't create any immediate harm.
- If the map contains any entry, the code will call `self.grow(0)` and start infinite probing in [this line](https://github.com/hyperium/http/blob/9c05e391e00474abaa8c14a86bcb0fc5eff1120e/src/header/map.rs#L1623).

Another problem is that the assertion for `MAX_SIZE` doesn't exist here, so it is possible to grow the map larger than `MAX_SIZE`.

[Demonstration](https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=4c45d68c59cf71191983b691d44e4595)

seanmonstar:
Fixes #352

### DIFF
--- a/src/header/map.rs
+++ b/src/header/map.rs
@@ -638,6 +638,8 @@ impl<T> HeaderMap<T> {
 
         if cap > self.indices.len() {
             let cap = cap.next_power_of_two();
+            assert!(cap < MAX_SIZE, "header map reserve over max capacity");
+            assert!(cap != 0, "header map reserve overflowed");
 
             if self.entries.len() == 0 {
                 self.mask = cap - 1;
--- a/tests/header_map.rs
+++ b/tests/header_map.rs
@@ -38,6 +38,22 @@ fn smoke() {
 }
 
 #[test]
+#[should_panic]
+fn reserve_over_capacity() {
+    // See https://github.com/hyperium/http/issues/352
+    let mut headers = HeaderMap::<u32>::with_capacity(32);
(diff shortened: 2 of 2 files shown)


src/header/map.rs
```diff
@@ -638,6 +638,8 @@ impl<T> HeaderMap<T> {
 
         if cap > self.indices.len() {
             let cap = cap.next_power_of_two();
+            assert!(cap < MAX_SIZE, "header map reserve over max capacity");
+            assert!(cap != 0, "header map reserve overflowed");
```
hawkw: TIOLI: maybe [`usize::checked_next_power_of_two`](https://doc.rust-lang.org/std/primitive.usize.html#method.checked_next_power_of_two) + `expect` would express intent more clearly? Not a blocker.

seanmonstar: I find the asserts clearer, myself.

Issue #355, opened 2019-11-16
`HeaderMap::Drain` can cause data race


Qwaz:
https://github.com/hyperium/http/blob/9c05e391e00474abaa8c14a86bcb0fc5eff1120e/src/header/map.rs#L2099-L2102

`HeaderMap::Drain` implements `Iterator` trait with `Item = (HeaderName, ValueDrain<'a, T>)`. This definition makes it possible to create multiple mutable references to the underlying `HeaderMap` by creating multiple `ValueDrain`, because there is no relation between the lifetime of `&mut self` and the returned `ValueDrain`.

This bug may invalidate `ValueDrain` iterators or make them return incorrect results. In multi-threaded scenario, the bug allows multiple threads to mutate underlying `Vec` at the same time, which I believe exploitable. However, I expect no code is sending `ValueDrain` to the other thread in practice.

[Demonstration](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f61a445dd5dca1f5516b540f74330d8b)

seanmonstar:
Yea, the call to `extra_values.swap_remove` isn't synchronized, so that's a problem. I think the solutions both have downsides, but I'll list them out:

- **Remove the `impl Send for ValueDrain`**, but this is an API breaking change. I suspect this should be incredibly uncommon. Perhaps the only sort of way someone would be affected is if using `rayon::par_iter` or similar with this.
- Put a `Mutex` into the `ValueDrain`. This negatively affects everyone for such a rare occurrence.
- **Re-write the internals of `HeaderMap` to not store all extra values in a single `Vec`.** This seems like the worst option, as it will take longer to do, and also negatively impacts memory usage and fragmentation.

PR #361, opened 2019-11-26
Remove Send from ValueDrain iterator

seanmonstar:
As noted in #355, the `ValueDrain` iterator has un-synchronized mutation of the `extra_values` `Vec`, and using a `Drain`, a user can get multiple `ValueDrain`s at the same time and use them in scoped threads.

This change removes the `Send` and `Sync` traits from `ValueDrain`. While technically a breaking change, it's expected to not affect anyone, and is the least-bad solution.

### DIFF
--- a/src/header/map.rs
+++ b/src/header/map.rs
@@ -3030,8 +3030,10 @@ impl<'a, T> Drop for ValueDrain<'a, T> {
     }
 }
 
-unsafe impl<'a, T: Sync> Sync for ValueDrain<'a, T> {}
-unsafe impl<'a, T: Send> Send for ValueDrain<'a, T> {}
+// `ValueDrain` cannot be `Send` because it has un-synchronized mutation of
+// the `extra_values` vector.
+//
+// See https://github.com/hyperium/http/issues/355
 
 // ===== impl RawLinks =====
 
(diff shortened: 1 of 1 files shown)


seanmonstar:
The point is to fix technically unsound behavior, without forcing user's to deal with all the other breaking changes.

hawkw:
yeah, i just wasn't sure if it was a concern that this is _technically_ a semver violation. seems fine to me 👍

carllerche:
What was the reason to opt for this instead of move values eagerly into the iterator?

seanmonstar:
Moving them eagerly adds a cost of allocation even if you're not going to use them, just to save a super obscure case.

carllerche:
@seanmonstar it could be "optimized" to be a small vec, then the allocation is avoided in most cases.

seanmonstar:
Is all that fiddling worth not making this change? Do we expect even a single person to be sending `ValueDrain` to another thread?

carllerche:
Making it `!Send` will break any usage w/ Tokio & calling `.await` while holding to the iterator.

seanmonstar:
Bleck, gross. I've open https://github.com/hyperium/http/pull/362 with a different solution.

PR #364, opened 2019-11-26
Update master to 0.2.x

seanmonstar:


Cargo.toml
```diff
@@ -4,7 +4,7 @@ name = "http"
 # - Update html_root_url in lib.rs.
 # - Update CHANGELOG.md.
 # - Create git tag
-version = "0.1.20"
+version = "0.2.0-a.0"
```
hawkw: I've noticed the general convention seems to be spelling this out, as in
```suggestion
version = "0.2.0-alpha.0"
```

PR #367, opened 2019-12-02
Revert "Refactor Method::is_idempotent"

sfackler:
Reverts hyperium/http#348

is_idempotent currently incorrectly returns false for PUT and DELETE.


XAMPPRocky:
My bad, I've opened #368 that contains a fix.

seanmonstar:
Fixed and released.

PR #368, opened 2019-12-02
Fix is_idempotent

XAMPPRocky:
Whoops, I forgot that an match guard applies to whole arm and not the branch. This PR fixes that.

### DIFF
--- a/src/method.rs
+++ b/src/method.rs
@@ -207,8 +207,8 @@ impl Method {
     /// more words.
     pub fn is_idempotent(&self) -> bool {
         match self.0 {
-            Put | Delete | _ if self.is_safe() => true,
-            _ => false
+            Put | Delete => true,
+            _ => self.is_safe(),
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/method.rs
```diff
@@ -207,8 +207,9 @@ impl Method {
     /// more words.
     pub fn is_idempotent(&self) -> bool {
         match self.0 {
-            Put | Delete | _ if self.is_safe() => true,
-            _ => false
+            Put | Delete => true,
+            _ if self.is_safe() => true,
```
sfackler: This can just be `m => m.is_safe()`

XAMPPRocky: The `is_safe` method is only implemented on `Method` not on `Inner` so changing to this gives method not found error.

sfackler: `_ => self.is_safe()` then.

seanmonstar:
If we didn't have tests that noticed the first time, we should add some :D

XAMPPRocky:
@seanmonstar Added tests.

PR #369, opened 2019-12-02
Remove Bytes from the public API

### LINKED ISSUE #366: Removing Bytes from the public API
We'd like to disconnect the stability if the `http` crate from the `bytes` crate. However, *using* `Bytes` inside `http`s types facilitates improved performance by reducing copies (especially as headers and URIs can be quite long in the real world).

This proposes a way forward without sacrificing either case:

- Remove all public APIs that uses `Bytes`. This includes `from_shared`, `TryFrom` impls, and `into_bytes` methods.
- Add `from_maybe_shared<T: AsRef<[u8]> + 'static>` constructors to types that would most benefit. These constructors could downcast the value into a `Bytes` if the versions match, and perform a copy if they don't. While a copy not caught at compile-time is unfortunate, the alternatives seem worse.

seanmonstar:
This should remove `Bytes` from the public API while it is unstable, and adds `from_maybe_shared` generic constructors that can allow improved performance (less copies) if the correct version of `Bytes` is used.

Closes #366

### DIFF
--- a/src/uri/scheme.rs
+++ b/src/uri/scheme.rs
@@ -5,7 +5,7 @@ use std::str::FromStr;
 
 use bytes::Bytes;
 
-use super::{ErrorKind, InvalidUri, InvalidUriBytes};
+use super::{ErrorKind, InvalidUri};
 use crate::byte_str::ByteStr;
 
 /// Represents the scheme component of a URI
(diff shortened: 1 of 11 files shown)


seanmonstar:
cc @LucioFranko @sfackler

sfackler:
Is the idea that the next release of `http` will be 1.0?

seanmonstar:
I'm strongly considering it, or otherwise a 0.2 meant as a preview of 1.0 just to check the new changes.

src/uri/scheme.rs
```diff
@@ -38,6 +38,7 @@ impl Scheme {
         inner: Scheme2::Standard(Protocol::Https),
     };
 
+    /*
```
LucioFranco: did you mean to keep this around?

LucioFranco:
I'm also happy to go 1.0 with http

PR #375, opened 2019-12-19
feat: support conversions from `url::Url`

davidbarsky:
## Motivation

- A few AWS services (like SQS, IAM, EC2) use `application/x-www-form-urlencoded`-based RPC calls. While this encoding is typically handled by the SDKs, this code needs to exist somewhere.
- An off-by-default feature flag enabling a `TryFrom` conversion from `url::Url` allows for building complex queries using `url::Url`'s builder methods, but then converting to `http::Uri` at request construction time.
- Since all of `http`'s builders implement accept a `TryFrom<T>` (in `http::request::Builder`'s case, [a `Uri: TryFrom<T>`](https://docs.rs/http/0.2.0/http/request/struct.Builder.html#method.uri)), this feature allows for passing a `url::Url` into a request builder.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,10 +21,15 @@ keywords = ["http"]
 categories = ["web-programming"]
 edition = "2018"
 
+[features]
+default = []
+url = ["url_2"]
+
 [dependencies]
 bytes = "0.5"
 fnv = "1.0.5"
 itoa = "0.4.1"
+url_2 = { package = "url", version = "2.1.0", optional = true }
 
 [dev-dependencies]
 indexmap = "1.0"
--- a/README.md
+++ b/README.md
@@ -66,6 +66,18 @@ fn main() {
 }
 ```
 
+`http` also supports conversions from
+[`url::Url`](https://github.com/servo/rust-url) using
+`std::convert::TryFrom`. To make use of this feature enable the `url`
+feature on `http`:
+
+```toml
+[dependencies]
+http = { version = "0.2", features = ["url"] }
+```
+
+Note that `http::Uri` is probably sufficient for most use-cases.
+
 # License
 
 Licensed under either of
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -21,10 +21,15 @@ keywords = ["http"]
 categories = ["web-programming"]
 edition = "2018"
 
+[features]
+default = []
+url = ["url_2"]
```
sfackler: I'd probably just have the feature called `url_2` which is already implied by the optional dependency.

davidbarsky:
(cc: @seanmonstar.)

Cargo.toml
davidbarsky: Errors out with: `Features and dependencies cannot have the same name: `url_2``. The two need to be different names.

sfackler: Optional dependencies are automatically features, you should just be able to remove the features block entirely.

davidbarsky: Oh, neat. I didn't realize that. Fixing.

davidbarsky: done.

seanmonstar:
Relevant discussion about why this hasn't been implemented in the past: https://github.com/hyperium/http/issues/73

davidbarsky:
@seanmonstar Thanks for the background context. There are a few circumstances—namely, the AWS use-case—where I _am_ constructing full URLs, not just URI fragments that compose into a target of an HTTP request. I think this PR implements the relatively conservative conversions as outlined in Carl's comment: https://github.com/hyperium/http/issues/73#issuecomment-319422772.

davidbarsky:
Not sure if any decision have been made on this PR but I think something along the lines of https://github.com/hyperium/http/pull/276 would address some of the goals of this PR. I'd be happy to revive #276.

(cc: @carllerche.)

PR #380, opened 2020-01-03
Have Status::as_str return static reference

upsuper:


upsuper:
@dekellum Hey! I saw that #381 is closed (not sure what happened there). What about this PR now? Should I rebase or?

dekellum:
CI was restored in a different way (#387), so you might try to merge with master to get this PR passing.

upsuper:
Thanks!

upsuper:
So... the CI is still broken.

I'd suggest you mark nightly with allow_failures, especially given that you have `deny(warnings)` which is very easy to be broken.

upsuper:
@dekellum could you have another look? it seems that I can't re-request review, and I have no idea why.

dekellum:
Your rebased change continues to look good to me!  

But I have no commit rights here, nor do I endorse, much less understand, its "governance".  All resent commits are by @seanmonstar and @carllerche was previously active.  You might try requesting a review from them?

PR #383, opened 2020-01-09
Add 'X-Forwarded-' headers

malthe:
Add headers `x-forwarded-for`, `x-forwarded-port`, and `x-forwarded-proto`, commonly used in load balancer setups.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+* Add headers `x-forwarded-for`, `x-forwarded-port`, and `x-forwarded-proto`, commonly used in load balancer setups.
+
 # 0.2.0 (December 2, 2019)
 
 * Add `Version::HTTP_3` constant.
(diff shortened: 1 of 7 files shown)


malthe:
The build fails on _nightly_ for an unrelated reason:

image

robjtede:
If this PR is updated, I can see this being somewhat [useful for web frameworks](https://github.com/actix/actix-web/blob/3f03af1c5928e1c4ea0cfab4d2ddfc7043b571f1/actix-web/src/info.rs#L16-L21) but my purist side says we should encourage the use of the more modern and already included [`Forwarded` header](https://github.com/actix/actix-web/blob/3f03af1c5928e1c4ea0cfab4d2ddfc7043b571f1/actix-web/src/info.rs#L82).

As shown, it is relatively easy to create these non-standard static/const headers using once_cell.

malthe:
I think the X-versions of the header are so widespread that it is perhaps beyond repair, see for example https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#defaults.

But that said, I suppose there's no better time than today to migrate to a standard setup with `Forwarded`.

seanmonstar:
I'd also prefer to suggest using the from_static constructor instead of adding these new constants...

PR #386, opened 2020-01-24
HeaderMap: Store pos and hash as u16

mbrubeck:
HeaderMap currently truncates all hashes to 16 bits, and limits its 
capacity to 32768 elements. The comments say this is done in order to store
positions and hashes as `u16`, to improve cache locality.  However, they are
currently stored as `usize`.  This branch changes the code to match the
comments.

This does not appear to cause any measurable improvement or regression in
speed in the HeaderMap benchmarks on my workstation. However, it should at
least reduce the memory footprint of every Headermap.

### DIFF
--- a/src/header/map.rs
+++ b/src/header/map.rs
@@ -232,7 +232,7 @@ enum Cursor {
 /// You may notice that `u16` may represent more than 32,768 values. This is
 /// true, but 32,768 should be plenty and it allows us to reserve the top bit
 /// for future usage.
-type Size = usize;
+type Size = u16;
 
 /// This limit falls out from above.
 const MAX_SIZE: usize = (1 << 15);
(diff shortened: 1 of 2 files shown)


seanmonstar:
Oh sweet!

(I've filed https://github.com/hyperium/http/pull/387 for the deprecation errors.)

mbrubeck:
Note that the `Link` and `Links` types still store indices as `usize`.  Changing these to u16 would probably be beneficial for some use cases, though they are not as important as `Size` and `HashValue` which together make up the entire `HeaderMap::indices` vector.

I didn't change `Link` and `Links` because it would touch a lot more code and add a lot more `u16`/`usize` conversions. This could be error-prone if not done carefully. However, it might be a good follow-up task.

PR #389, opened 2020-01-27
Allow nightly build to fail

upsuper:
It's a common practice to allow failures in nightly for CI, as nightly is changing in a faster pace than most other projects, and it would be painful that you need to fix something whenever you want to change something.

This is especially a problem given that this project has `#[deny(warnings)]` which can be even more fluid.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ sudo: false
 cache: cargo
 
 matrix:
+  allow_failures:
+    - rust: nightly
   include:
     - rust: stable
     - rust: beta


seanmonstar:
I think it's preferable to keep the nightly failure, so that way we know if there's a bug in nightly. In case where I've allowed this failure, I often don't notice until nightly is promoted to beta.

carllerche:
I preference is to pin nightly to a specific version and to periodically increment it.

I would say either pin nightly **or** remove the nightly build. :-1: on allowing to fail.

PR #409, opened 2020-04-03
[src/error.rs] pub ErrorKind

SamuelMarks:
I'm confused how to percolate errors from `http` through my library.

I want to percolate up a [`http::uri::ErrorKind::InvalidFormat`](https://github.com/hyperium/http/blob/c175fea/src/uri/mod.rs#L128-L141) if [`.host()`](https://github.com/hyperium/http/blob/c175fea/src/uri/mod.rs#L552-L555) is [`None`](https://doc.rust-lang.org/src/core/option.rs.html#159).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ version = "0.2.1"
 readme = "README.md"
 documentation = "https://docs.rs/http"
 repository = "https://github.com/hyperium/http"
-license = "MIT/Apache-2.0"
+license = "Apache-2.0 OR MIT"
 authors = [
   "Alex Crichton <alex@alexcrichton.com>",
   "Carl Lerche <me@carllerche.com>",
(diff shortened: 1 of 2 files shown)


dekellum:
The private `http::error::ErrorKind` shouldn't be in your way, since `http::Error` has `is` and `get_ref` for read access, and implements `From<InvalidUri>` (if "percolate" includes creating them).  However, I see that `http::uri::InvalidUri` is itself opaque: not user constructable and only offers a description to read.  Is that the real issue?

The prominent style here AFAICT, is to never expose a public `enum`. This unfortunately tends to leave users with only a description string, but #303 demonstrates willingness to improve usability in others ways. Perhaps kind-specific constructor functions and `is_<kind>(&self) -> bool` (e.g. `is_invalid_format`) methods could be added to `InvalidUri`?

SamuelMarks:
@dekellum So how do I create a `http::uri::InvalidUri` in my codebase?

SamuelMarks:
@seanmonstar Why was this closed? - I have a genuine issue + fix here

Issue #412, opened 2020-04-11
Document and test the use of unsafe


sbosnick:
A recent Medium [article](https://medium.com/@shnatsel/smoke-testing-rust-http-clients-b8f2ee5db4e6) criticized 3 of the 7 rust http clients it reviewed because of their dependency on this crate and because of this crate's use of `unsafe`. The criticism does not appear to be founded (i.e. it claims that  `HeaderMap` is a reimplementation of `HashMap` without explaining how to use `HashMap` as a multimap) and the author's tests of these 3 http clients did not show any exploitable problems related to this crate. The article appears to an example of withoutboats' [post](https://boats.gitlab.io/blog/post/vulnerabilities/) about allegations of vulnerabilities in Rust libraries.

Despite the limitations of the Medium article, `http` is (and is designed to be) a high-level dependency of many network exposed crates so the sound use of `unsafe` is important. The use of `unsafe` has already been audited (see for example rust-secure-code/safety-dance#37). In my preliminary review of the use of `unsafe` I found that they were sound (with one possible exception I am still investigating). The soundness, though, relies on implicit invariants that, in some instance, are not local to the use of `unsafe`.

I propose to try to increase the robustness (in the sense described [here](https://accu.org/index.php/journals/1585)) of the use of `unsafe` in `http` by:

- Adding unit tests for edge condition involving `unsafe` code paths
- Refactoring the code involving `unsafe` into smaller leaf modules that present a safe wrapper around the unsafe code (without introducing performance regressions)
- Documenting the invariants that make the use of `unsafe` sound (making explicit what is implicit)
- Adding additional performance tests for uses-no-unsafe alternatives to `HeaderMap` (such as [multimap](https://crates.io/crates/multimap)) so that the performance trade-offs of the current implantation are more easily apparent

The goal isn't to reduce or eliminate the use of `unsafe` (though a reduction in in the number of uses of `unsafe` may be a by-product). It is rather to make the soundness of the use of `unsafe` easier for someone else to audit.

I plan to review the use of `unsafe` file by file as follows (the number in parenthesis is the number of occurrences of `unsafe` in the file):

- [ ] byte_str.rs (2)
- [ ] method.rs (2)
- [ ] uri/path.rs (1)
- [ ] uri/authority.rs (3)
- [ ] uri/mod.rs (3)
- [ ] header/value.rs (2)
- [ ] header/name.rs (11)
- [ ] header/map.rs (19)

[...]

PR #413, opened 2020-04-12
Audit the use of unsafe in uri/path.rs

sbosnick:
Reorganize the one function with a use of unsafe (from_shared()) to highlight that it does a scan of each byte that it eventually passes to ByteStr::from_utf8_unchecked(). This makes it apparent (as described in the "Safety" comment) that the input to from_utf8_unchecked() is valid UTF-8 and is, thus, sound.

This is a part of #412.

### DIFF
--- a/src/uri/path.rs
+++ b/src/uri/path.rs
@@ -518,6 +518,7 @@ mod tests {
         assert_eq!("a", pq_bytes(&[b'a']));
         assert_eq!("a?", pq_bytes(&[b'a', b'?']));
         assert_eq!("a?b", pq_bytes(&[b'a', b'?', b'b']));
+        assert_eq!("a?{b}", pq_bytes(&[b'a', b'?', b'{', b'b', b'}']));
         assert_eq!("a", pq_bytes(&[b'a', b'#', b'b']));
         assert_eq!("a?b", pq_bytes(&[b'a', b'?', b'b', b'#', b'c']));
 
@@ -530,6 +531,7 @@ mod tests {
     fn from_invalid_u8_slice_is_error() {
         assert!(is_invalid_pq_bytes(&[0xc0])); // invalid UTF-8
         assert!(is_invalid_pq_bytes(&[b' '])); // need percent encoding
+        assert!(is_invalid_pq_bytes(&[b'{'])); // need percent encoding
         assert!(is_invalid_pq_bytes(&[b'a', b'?', 0xc0])); // invalid UTF-8
         assert!(is_invalid_pq_bytes(&[b'a', b'?', b' '])); // needs percent encoding
     }
--- a/src/uri/path.rs
+++ b/src/uri/path.rs
@@ -54,6 +54,9 @@ impl PathAndQuery {
                     0x7B |
                     0x7D if query.is_some() => Ok((query, i+1)),
 
+                    // all bytes 0x80 and above match here (among others)
+                    // so all Ok() returns in the other match arms identify
+                    // a single byte UTF-8 code point
                     _ => Err(InvalidUri(ErrorKind::InvalidUriChar)),
                 }
             })?;
@@ -61,6 +64,9 @@ impl PathAndQuery {
         src.truncate(len as usize);
 
         Ok(PathAndQuery {
+            // Safety: The try_fold() checks that each byte in the now truncated
+            // src is a single byte UTF-8 code point so src as a whole is valid
+            // UTF-8.
             data: unsafe { ByteStr::from_utf8_unchecked(src) },
             query: query.unwrap_or(NONE),
         })
--- a/src/uri/path.rs
+++ b/src/uri/path.rs
@@ -19,74 +19,50 @@ const NONE: u16 = ::std::u16::MAX;
 impl PathAndQuery {
     // Not public while `bytes` is unstable.
     pub(super) fn from_shared(mut src: Bytes) -> Result<Self, InvalidUri> {
-        let mut query = NONE;
-        let mut fragment = None;
-
-        // block for iterator borrow
-        {
-            let mut iter = src.as_ref().iter().enumerate();
-
-            // path ...
-            for (i, &b) in &mut iter {
-                // See https://url.spec.whatwg.org/#path-state
-                match b {
-                    b'?' => {
-                        debug_assert_eq!(query, NONE);
-                        query = i as u16;
-                        break;
-                    }
-                    b'#' => {
-                        fragment = Some(i);
-                        break;
-                    }
+        let (query, len) = src.as_ref().iter()
+            // stop at the fragment specifier
+            .take_while(|&&c| c != b'#')
+            .try_fold((None, 0u16), |(query, i), &c| {
+                match c {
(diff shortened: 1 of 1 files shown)


src/uri/path.rs
```diff
@@ -511,7 +493,40 @@ mod tests {
...
+    #[test]
+    fn from_u8_slice_equates_with_string() {
+        assert_eq!("/", pq_bytes(&[b'/']));
+        assert_eq!("/ab", pq_bytes(&[b'/', b'a', b'b']));
+        assert_eq!("a", pq_bytes(&[b'a']));
+        assert_eq!("a?", pq_bytes(&[b'a', b'?']));
+        assert_eq!("a?b", pq_bytes(&[b'a', b'?', b'b']));
+        assert_eq!("a?{b}", pq_bytes(&[b'a', b'?', b'{', b'b', b'}']));
+        assert_eq!("a", pq_bytes(&[b'a', b'#', b'b']));
+        assert_eq!("a?b", pq_bytes(&[b'a', b'?', b'b', b'#', b'c']));
```
hawkw, line 505: May eventually be worth having property tests in addition, but definitely not a blocker.

src/uri/path.rs
```diff
@@ -19,74 +19,56 @@ const NONE: u16 = ::std::u16::MAX;
...
-
-        if let Some(i) = fragment {
-            src.truncate(i);
-        }
+        src.truncate(len as usize);
 
         Ok(PathAndQuery {
+            // Safety: The try_fold() checks that each byte in the now truncated
+            // src is a single byte UTF-8 code point so src as a whole is valid
+            // UTF-8.
```
hawkw, line 69: Nit, take it or leave it: I might split this into two sentences. Currently, it feels like a bit of a run-on.

Maybe:
```suggestion
            // Safety: The try_fold() checks that each byte in the (now truncated)
            // src is a single byte UTF-8 code point. Therefore,  src as a whole
            // is valid UTF-8.
```
or something?

sbosnick: Fixed.

src/uri/path.rs
sbosnick: I would be happy to add property tests but my thoughts are that that should be a separate pull request. If this is a good approach I could add an issue for this as a marker and come back to it later.

sbosnick:
If there are any concerns about a possible performance impact I would be happy to add benchmarks to answer that question.

src/uri/path.rs
```diff
@@ -19,74 +19,56 @@ const NONE: u16 = ::std::u16::MAX;
...
-                        break;
-                    }
-                    b'#' => {
-                        fragment = Some(i);
-                        break;
-                    }
+        let (query, len) = src.as_ref().iter()
+            // stop at the fragment specifier
+            .take_while(|&&c| c != b'#')
+            .try_fold((None, 0u16), |(query, i), &c| {
```
seanmonstar, line 25: Does passing these values along due to `fold` optimize differently? Do we even have benchmarks parsing URIs?

Also, this isn't a super strong feeling, but I kind of feel that the previous 2 loops is a little clearer that we're looking for two things....

sbosnick: On the second point, the two loop version emphasizes that we are looking for two things, the one loop version emphasizes that we are looking at every byte in `src` until the first `'#'`. The latter is what we need to confirm the soundness of the call the `ByteStr::from_utf8_unchecked()` later in the function. (This is just a shift in emphasis; the two loop version checks the same things.) It is possible to describe the 2 loop version with comments describing the loop control flow (how the early exits and the normal end-of-loop combine to ensure that every byte until the first `'#'` gets checked) but the one loop version accomplishes this in code, rather than comments. This is a trade-off on the code-quality issue.

Unfortunately, though, there is a performance regression. (When I checked quickly in the past it didn't look like there was but I checked it more closely after this comment and there is.) There are a few benchmarks for parsing URI's in `benches/uri.rs`. They don't benchmark all of the URI parsing code (for example they don't cover parsing the scheme or authority) but two of the benchmarks cover the path and query parsing code. These benchmarks show a performance regression in this code over that in the master branch.

I will try a version of the one loop approach with external iteration instead of `try_fold` to see if that addresses the performance issue.

sbosnick: The latest commit has resolved the performance regression (to within the margin of error).

sbosnick:
The latest commit removes the performance regression that the benchmarks in `benches/uri.rs` showed (to within the margin of error). This retains the one-loop version of the core parsing code but changed from internal iteration (using `try_fold()`) to external iteration.

It appears that the outstanding issue then is whether this code should use a two-loop version that emphasizes parsing for two separate things (the path and the query) of a one-loop version that emphasized checking each  byte in turn that will eventually be passed to `ByteStr::from_utf8_unchecked()`.

I favour the latter because it makes the soundness easier to see. I can convert this back to a two-loop version, though, if there is a reason to prefer that approach.

seanmonstar:
Thanks! I tried out the benchmarks with the newest changes (commit 
d4718c9 ), here's what I see:

Before:

```
test uri_parse_relative_medium ... bench:         110 ns/iter (+/- 0) = 545 MB/s
test uri_parse_relative_query  ... bench:         134 ns/iter (+/- 1) = 626 MB/s
test uri_parse_slash           ... bench:          40 ns/iter (+/- 1) = 25 MB/s
```

After:

```
test uri_parse_relative_medium ... bench:         132 ns/iter (+/- 1) = 454 MB/s
test uri_parse_relative_query  ... bench:         169 ns/iter (+/- 1) = 497 MB/s
test uri_parse_slash           ... bench:          41 ns/iter (+/- 0) = 24 MB/s
```

sbosnick:
The latest commit improves the performance of this branch to just shy of master. The results that I see on the uri benchmarks are as follows:

Master:

```
test uri_parse_relative_medium ... bench:         120 ns/iter (+/- 0) = 500 MB/s
test uri_parse_relative_query  ... bench:         170 ns/iter (+/- 0) = 494 MB/s
test uri_parse_slash           ... bench:          30 ns/iter (+/- 0) = 33 MB/s
```

https://github.com/hyperium/http/pull/413/commits/27ff0cb0bf1e2b41755c24e24d188cda63b4dc76:

```
test uri_parse_relative_medium ... bench:         126 ns/iter (+/- 1) = 476 MB/s
test uri_parse_relative_query  ... bench:         176 ns/iter (+/- 0) = 477 MB/s
test uri_parse_slash           ... bench:          30 ns/iter (+/- 0) = 33 MB/s
```

So the two benchmarks that are affected by these changes are 6 ns/iter slower as compared with master.

This change returns to a two-loop version (that optimizes better than the different variations I tried of the one-loop version) but it uses `take_while()` to end checking the bytes at the fragment specifier instead of having an early exit in each of the two loops. This makes it more obvious that between the two loops they check every byte up to the point where the later code truncates `src` (this is embedded in the code itself instead of having to be described in comments).

I don't know how crucial the uri parsing code is on the fast path of the crates that rely on `http`. I suggest, though, that if 6 ns/iter is an important difference then we need more (and better) benchmarks of the uri parsing (which I am prepared to write).

PR #414, opened 2020-04-15
Audit use of unsafe in uri/authority.rs

sbosnick:
Refactor the common code from three ways of creating an Authority into a private create_authority() function, which removes some code duplication (including duplicated uses of "unsafe"). Add comments describing the implicit preconditions and postconditions in create_authority() and the functions it calls which then make explicit the sound use of "unsafe" (as described in the "Safety" comment in create_authority().

This is part of #412.

### DIFF
--- a/src/uri/authority.rs
+++ b/src/uri/authority.rs
@@ -23,6 +23,8 @@ impl Authority {
 
     // Not public while `bytes` is unstable.
     pub(super) fn from_shared(s: Bytes) -> Result<Self, InvalidUri> {
+        // Preconditon on create_authority: trivially satisfied by the
+        // identity clousre
         create_authority(s, |s| s)
     }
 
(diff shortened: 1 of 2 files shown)


src/uri/authority.rs
```diff
@@ -23,15 +23,9 @@ impl Authority {
...
-        let authority_end = Authority::parse_non_empty(&s[..])?;
-
-        if authority_end != s.len() {
-            return Err(ErrorKind::InvalidUriChar.into());
-        }
-
-        Ok(Authority {
-            data: unsafe { ByteStr::from_utf8_unchecked(s) },
-        })
+        // Preconditon on create_authority: trivially satisfied by the
```
hawkw, line 26: typo:
```suggestion
        // Precondition on create_authority: trivially satisfied by the
```

sbosnick: Thank you for this. I have applied your suggestion.

sbosnick: Thank you for this. I have applied your correction.

PR #417, opened 2020-04-25
Audit use of unsafe in uri/mod.rs

sbosnick:
Added tests for attempts to parse various types of invalid `Uri`'s including ones with invalid UTF-8 bytes in them. Added a test for parsing `&[u8]` as a Uri where it has invalid UTF-8 bytes in the fragment. This test accepts the `Uri` as valid because `Uri` (currently) does not expose the fragment so those bytes are never interpreted as a `&str`.

Refactored the `parse_full()` function to eliminate some code duplication and thereby simplify the function. Finally, added comments to `parse_full()` and the functions it calls to document the postconditions that `parse_full()` relies on to make its use of `unsafe` sound.

This PR has a weak discrepancy on #414 and #416 in the sense that some of the comments added in this PR make more sense in light of the comments added in the earlier two PR's. There is no dependency in this PR on the earlier PR's to build or to run tests. 

This is a part of #412.

### DIFF
--- a/src/uri/mod.rs
+++ b/src/uri/mod.rs
@@ -804,6 +804,8 @@ fn parse_full(mut s: Bytes) -> Result<Uri, InvalidUri> {
             scheme.truncate(n);
 
             // Allocate the ByteStr
+            // Safety: the postcondition on Scheme2::parse() means that
+            // s[0..n+3] is valid UTF-8. scheme is a subslice of s[0..n+3].
             let val = unsafe { ByteStr::from_utf8_unchecked(scheme) };
 
             Scheme2::Other(Box::new(val))
@@ -828,6 +830,10 @@ fn parse_full(mut s: Bytes) -> Result<Uri, InvalidUri> {
 
     let authority = s.split_to(authority_end);
     let authority = Authority {
+        // Safety: The postcondition on Authority::parse() means that
+        // s[0..authority_end] is valid UTF-8 after that call. The call
+        // to s.split_to() means that authority here is what s[0..authority_end]
+        // was after the call to Authority::parse().
         data: unsafe { ByteStr::from_utf8_unchecked(authority) },
     };
 
--- a/src/uri/mod.rs
+++ b/src/uri/mod.rs
@@ -785,12 +785,14 @@ impl From<Uri> for Parts {
     }
 }
 
+// parse_full() parses a Uri that includes more than just a path. It
+// expects that at least one of the scheme or authority will be present
+// as well.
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -160,6 +160,9 @@
 
 #![deny(warnings, missing_docs, missing_debug_implementations)]
 
+// Nightly Regression https://github.com/rust-lang/rust/issues/70814
+#![allow(unknown_lints, unused_braces)]
```
sbosnick, line 164: The issue that was causing the nightly build to fail (issue 70814) may have been fixed (or at least have had a work around added to rustc) in nightly-2020-04-26. We should remove this change before merging to see if nightly is still broken.

sbosnick: I reverted this change and the pull request has built on nightly so this build breakage is resolved.

src/uri/mod.rs
```diff
@@ -813,28 +817,23 @@ fn parse_full(mut s: Bytes) -> Result<Uri, InvalidUri> {
...
         }
-
-        let authority = Authority {
-            data: unsafe { ByteStr::from_utf8_unchecked(s) },
-        };
-
-        return Ok(Uri {
-            scheme: scheme.into(),
-            authority: authority,
-            path_and_query: PathAndQuery::empty(),
```
seanmonstar, line 827: This used to just assume `empty()`, and now its calling `parse(s)?`, right?

sbosnick: Not exactly. It used to assume `PathAndQuery::empty()`  and now it is calling `PathAndQuery::from_shared(s)` where `s` is an empty `Bytes`. We know that `s` is an empty `Bytes` because we test that `authority_end` equals the whole length of `s` at line 821 (of the changed file) and then call `s.split_to(authority_end)` at line 831. This latter call becomes (in this case) `s.split_to(s.len())` which results in `s` being left empty.

`PathAndQuery::empty()` and `PathAndQuery::from_shared(s)` where `s` is empty both produce a `PathAndQuery` with `data` the equivalent of `ByteStr::new()` and `query` as `NONE` (the sentinel value declared in `uri::path`).

Logically this change still produces the same result in the `scheme.is_none()` case.

Performance wise, this particular case isn't covered by any of the existing benchmarks in `benches/uri.rs`. Although `PathAndQuery::from_shared(s)` in general is linear in the size of `s`, when `s` is empty this a constant operation, though this doesn't answer the performance question without having benchmarks.

If there are any concerns about the performance implications of this change I would be happy to add a benchmark to `benches/uri.rs` to cover this case.

PR #419, opened 2020-05-03
Audit the use of unsafe in header/value.rs

sbosnick:
Reorganized the `Debug` impl for `HeaderValue` to make the invariants that its use of `unsafe` rely upon for soundness more obvious. This also adds a few more test cases for edges cases of the `Debug` impl. This also adds comments to all of the uses of `unsafe` in header/value.rs to explain and document their soundness.

One of the uses of `unsafe` in this file is in the function [`from_maybe_shared_unchecked()`](https://docs.rs/http/0.2.1/http/header/struct.HeaderValue.html#method.from_maybe_shared_unchecked) which is part of the public API of this crate. Currently this function is an `unsafe` function, but if `unsafe` functions are intended to signify caller responsibility for preventing undefined behaviour then this function is safe. There are no parameters you can pass to `from_maybe_shared_unchecked()` that will cause undefined behaviour.

There are, however, parameters you can pass to `from_maybe_shared_unchecked()` that would create a `HeaderValue` that did not comply with [RFC 7230 (section 3.2)](https://tools.ietf.org/html/rfc7230#section-3.2). That is, a `HeaderValue` that contained (for example) a NUL byte (0x00). If `unsafe` functions are intended to signify caller responsibility for maintaining some precondition (not just avoiding undefined behaviour) then it may make sense to keep `from_maybe_shared_unchecked()` as an `unsafe` function.

Removing `unsafe` from `from_maybe_shared_unchecked()` is an API change but it does not appear to be a breaking change.

Before this is merged we have to make a decision on the approach to `unsafe` functions to use and may further changes to `from_maybe_shared_unchecked()` (I can at least improve the doc comments if the `unsafe` should remain).

This is part of #412.

### DIFF
--- a/src/header/value.rs
+++ b/src/header/value.rs
@@ -346,24 +346,18 @@ impl fmt::Debug for HeaderValue {
             f.write_str("\"")?;
             let bytes = self.as_bytes();
 
-            // Invariant1: bytes[range] is valid UTF-8 (for both instanses of range).
-            // Invariant2: In the try_fold() closure bytes[range.end] == b.
+            // Invariant: bytes[range] is valid UTF-8 (for both instanses of range).
             //
             // bytes[0..0] is empty so the starting value of range trivally satisfies
-            // the invariant 1. The first invocaton of the closure has range.end == 0
-            // and b == bytes[0], satisfying the second invariant.
-            //
-            // On the nth invocation of the closure b = bytes[range.end]. The end of
-            // range returned from the closure is range.end+1. On the n+1th invocation
-            // of the closure b will be the next byte of bytes which is bytes[range.end+1].
-            // The second invariant follows by induction.
-            let range = bytes.iter().try_fold(0..0, |range, &b| {
+            // the invariant.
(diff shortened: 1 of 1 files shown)


sbosnick:
The latest commits (specifically https://github.com/hyperium/http/pull/419/commits/6364f795fb860e959d7133688f2e797be148ae57) make an opinionated change to marking `from_maybe_shared_unchecked` as an `unsafe` function. Specifically it removes the `unsafe` from the function definition.

As described in more detail in in the commit message for https://github.com/hyperium/http/pull/419/commits/6364f795fb860e959d7133688f2e797be148ae57, this is both sound and is not a breaking change.

This change is taking the design view that `unsafe` function in the public API should be restricted to functions where the caller bears responsibility for preventing undefined behaviour and not as a general means of indicating caller responsibility for ensuring that API constraints unrelated to undefined behaviour are met.

I am making this change to have a concrete proposal for consideration. If there are reasons to prefer the other approach to `unsafe` functions in the public API then I can change this back and explain the constraints (and their relationship to avoiding undefined behaviour) through comments.

PR #420, opened 2020-05-07
Replace "cfg(test)" with "cfg(doctest)" for readme testing

GuillaumeGomez:
Rustdoc now passes "doctest" when running in test mode.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -160,11 +160,11 @@
 
 #![deny(warnings, missing_docs, missing_debug_implementations)]
 
-#[cfg(test)]
+#[cfg(doctest)]
 #[macro_use]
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -19,7 +19,7 @@ jobs:
           - stable
           - beta
           - nightly
-          - 1.39.0
```
seanmonstar, line 22: This is the MSRV.

GuillaumeGomez: I can add a version detection for test builds and if too low, I will use "disable-doctest" feature (or another feature name) so you won't have build failures on versions lower than 1.40. What do you think?

PR #422, opened 2020-05-19
feature: make possible to set port number in Uri::builder

Dushistov:
Make possible to set port number in `Uri::builder`.
I work with external API that gives me:
        - scheme;
        - hostname;
        - port;
        - path; 
It is inconvenient to use format/write macros instead of just call builder method.
ref #206

### DIFF
--- a/src/uri/builder.rs
+++ b/src/uri/builder.rs
@@ -1,7 +1,10 @@
-use std::convert::{TryFrom, TryInto};
+use std::{
+    convert::{TryFrom, TryInto},
+    fmt::Write,
+};
 
-use super::{Authority, Parts, PathAndQuery, Scheme};
-use crate::Uri;
+use super::{Authority, ErrorKind, InvalidUriParts, Parts, PathAndQuery, Scheme};
+use crate::{byte_str::ByteStr, Uri};
 
 /// A builder for `Uri`s.
 ///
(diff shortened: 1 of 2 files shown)


src/uri/mod.rs
```diff
@@ -110,6 +110,9 @@ pub struct Parts {
     /// The authority component of a URI
     pub authority: Option<Authority>,
 
+    /// The port component of a URI
+    pub port: Option<u16>,
```
seanmonstar, line 114: The port is part of `authority`, we probably shouldn't have it in a two places.

src/uri/builder.rs
```diff
@@ -78,6 +78,14 @@ impl Builder {
         })
     }
 
+    /// Set the port number for URI, will be part of `Authority`
+    pub fn port(self, port: u16) -> Self {
+        self.map(move |mut parts| {
```
seanmonstar, line 83: I'd move the conversion to here, instead of adding a new field on `Parts`.

src/uri/builder.rs
```diff
@@ -78,6 +78,14 @@ impl Builder {
         })
     }
 
+    /// Set the port number for URI, will be part of `Authority`
+    pub fn port(self, port: u16) -> Self {
```
seanmonstar, line 82: This could probably be generic over something that can be converted into `Port`, so a user can easily pass `3000`, or `authority.port()`, etc.

src/uri/mod.rs
Dushistov: I am care about different order: `.port().authority()`,
but looks like `Uri::Builder` do not handle the misuse like `.authority().authority()`,
and so on. So I made changes as you suggest, so in new variant 
`.port().authority()` authority just overwrite what `port` is done.

src/uri/builder.rs
Dushistov: Done

src/uri/builder.rs
Dushistov: I can not see how to be generic over `Port` and `u16`, so I added `port_u16`.

Dushistov: Fixed

Dushistov:
@seanmonstar  I fixed all issues that you mentioned.

Dushistov:
@seanmonstar,  friendly ping.

Dushistov:
@seanmonstar, friendly ping!

Dushistov:
@seanmonstar, friendly ping!

Dushistov:
Rebase against current master

Dushistov:
Close in favor of #496

PR #428, opened 2020-05-31
Audit use of unsafe in header/name.rs

sbosnick:
This PR documents and improves the use of `unsafe` in the `header::name` module. That module has two kinds of usages of `unsafe`: the use of uninitialized memory and calls to `ByteStr::from_utf8_unchecked()`.

The uninitialized memory in `header::name` is used as a temporary buffer when normalizing the bytes that represent a name before matching them against the standard header names. This PR upgrades the uninitialized memory from the deprecated  [`mem::unititialized()`](https://doc.rust-lang.org/stable/std/mem/fn.uninitialized.html) to an array of [`MaybeUninit<u8>` ](https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html). This upgrade has the useful side effect of mostly localizing the use of `unsafe` related to uninitialized memory to the two implementations of `parse_hdr()`.  

The PR documents (through comments) the preconditions and post-conditions that make the use of uninitialized memory sound. The performance sensitive nature of `parse_hdr()` make these comments harder to follow than they might have been had it been possible to refactor parse_hdr() without causing a performance regression (I started such a refactoring but backed off when it caused regressions). The documented preconditions and post-conditions show the reasoning on how each `MaybeUninit<u8>` is initialized before it is accessed for reading and before it is used as a part of the return from `parse_hdr()`.

The PR also document the preconditions, post-conditions, and invariant that make the use of `unsafe` related to `ByteStr::from_utf8_unchecked` sound. These comments are also more extensive than they might have been because of the performance sensitive nature of `parse_hdr()`.

Finally the PR adds a new set of micro-benchmarks related to parsing the standard header names.

The new benchmarks make use of the [criterion](https://crates.io/crates/criterion) micro-benchmarking framework. This adds a new dev-dependency which is a major downside of this change. I suggest, though, that this downside is offset by the following upsides:
- criterion allows for running benchmarks on stable
- criterion makes it easy to run performance comparisons with either the last benchmark run or with a named baseline (and reports statistically significant improvements or regressions)
- criterion allows for more reliable benchmarks because of its separate warmup and measuring phases
- criterion is a well written, well designed, and actively maintained crate

This is part of #412.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ seahash = "3.0.5"
 serde = "1.0"
 serde_json = "1.0"
 doc-comment = "0.3"
+criterion = "0.3.2"
 
 [[bench]]
 name = "header_map"
@@ -42,6 +43,11 @@ path = "benches/header_map/mod.rs"
 [[bench]]
 name = "header_name"
 path = "benches/header_name.rs"
+
+[[bench]]
+name = "header_name2"
+path = "benches/header_name2.rs"
+harness = false
 
 [[bench]]
 name = "header_value"
--- a/benches/header_name2.rs
+++ b/benches/header_name2.rs
@@ -1,0 +1,52 @@
+use criterion::{criterion_group, criterion_main, BenchmarkId,Criterion, Throughput};
+use http::header::HeaderName;
+
+// This is a list of some of the standard headers ordered by increasing size.
+// It has exactly one standard header per size (some sizes don't have a standard
+// header).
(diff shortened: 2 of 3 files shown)


saethlin:
@seanmonstar It's been a little while since this PR was opened. Is there anything community members can do to move this PR forward?

LucioFranco:
I think it makes sense to merge https://github.com/hyperium/http/pull/499 then apply this patch on top since a lot of code was removed in #499 that was touched here.

seanmonstar:
Merged in #555 (thanks so much!)

PR #436, opened 2020-09-02
Make StatusCode::as_str return 'static str

e00E:
In the current version the lifetime is tied to self which is needlessly
restrictive as the strings are in fact static.


dekellum:
See also #380

kaj (approved):
Nice and simple change, loooks good to me.

PR #438, opened 2020-10-18
Make StatusCode allow 600-999

quininer:
Also change `StatusCode` to `NonZeroU16`, I think this will help performance.

see https://rust.godbolt.org/z/1Evhbx


src/status.rs
```diff
@@ -85,12 +88,14 @@ impl StatusCode {
...
-        if a == 0 || a > 5 || b > 9 || c > 9 {
+        if a == 0 || a > 9 || b > 9 || c > 9 {
             return Err(InvalidStatusCode::new());
         }
 
         let status = (a * 100) + (b * 10) + c;
-        Ok(StatusCode(status))
+        NonZeroU16::new(status)
+            .map(StatusCode)
+            .ok_or_else(InvalidStatusCode::new)
```
jplatte, line 98: Looks like the allows (now as well as before) status codes < 100, should probably just be

```suggestion
        StatusCode::from_u16(status)
```

quininer: see L91

jplatte: Oh, I missed the `a == 0` part. Nevermind.

Matthias247:
You should describe the use-case for this. The addition of status codes with printable representations will increase the binary (rough estimate: 400 status codes * (3 bytes len + 8 bytes length indicator + align = 16bytes) = 6kB).
So this shouldn't be done unless there is also a reason for it.

I guess one could be proxying, where an upstream return an arbitrary 3-digit number, which is not bounded up to 599?

quininer:
Many examples have been cited in https://github.com/hyperium/http/issues/144 . About size, I will try to make it compact. (3 * 900) = 2700, this will be smaller than it is now.

src/status.rs
```diff
@@ -126,7 +131,14 @@ impl StatusCode {
     /// ```
     #[inline]
     pub fn as_str(&self) -> &str {
-        CODES_AS_STR[(self.0 - 100) as usize]
+        let offset = (self.0.get() - 100) as usize;
```
Matthias247, line 134: Btw: Since the method above is defined  as `&str` the implementation is actually not expected to return a `&'static str` at the moment. So we could change it if concerned about storage with more status code. 

However I'm not sure about a good way to exploit this property. Embed the 3 bytes inside `StatusCode`? That makes that one bigger. Formatting it into thread-local storage might not work, due to this having to be `Sync`.

I would probably just have returned a fixed-length string, but that would be a breaking change now.

geota:
@quininer  @Matthias247 is there an update to getting this merged? I have a vested interest to get this merged and could try to pick this up if @quininer is unable to finish off the PR, but I don't want to step on any toes.

seanmonstar:
I'd tentative consider merging, putting aside my strong feeling that servers shouldn't do this.

I think the issue of added binary size is valid. I don't know a great way to fix that, perhaps `StatusCode::as_str` should go away completely?

quininer:
I think size issue should not prevent this PR, because this PR has made size smaller than before.

geota:
@seanmonstar @Matthias247 can we figure out what the next steps on this PR should be? Is size an issue? @quininer made the size smaller than before, so I feel that any other improvements here should be left to a future PR.

quininer:
Close by https://github.com/hyperium/http/pull/443

PR #442, opened 2020-11-07
Implement a non-failing Response Builder.

### LINKED ISSUE #330: Replace error-returning builders with typestate-based builders
It'd be really handy to replace the current builders with a set of builders for Request, Response, and URI that track the builder's state using some sort of typestate, allowing for compile-time checks that a Request/Response/URI were correctly built.

(Following up from Twitter: https://twitter.com/seanmonstar/status/1149772057212669952)

kaj:
Fixes one part of #330.  Creating a Response with the builder pattern should not require handling errors unless opting in by using a `try_`-method (which returns a `Result<Builder2>`, not a `Builder2` with an internal error state).

This is really a breaking change to `response::Builder`, but rather than suggesting a breaking change directly, I introduce an alternative `response::Builder2`.  If this is accepted, the original `Builder` is deprecated, and in a later release `Builder2` can be renamed back to `Builder`.

### DIFF
--- a/src/response.rs
+++ b/src/response.rs
@@ -210,6 +210,17 @@ pub struct Builder {
     inner: Result<Parts>,
 }
 
+/// An HTTP response builder
+///
+/// This type can be used to construct an instance of `Response` through a
+/// builder-like pattern.
+///
+/// This variant avoids error states.
+#[derive(Debug)]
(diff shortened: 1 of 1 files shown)


kaj:
Replaced by #483 .

PR #445, opened 2020-11-22
Return error if multiple brackets exist in the authority

### LINKED ISSUE #435: Unexpected panic for malformed URI with brackets
Doing some fuzzing I encountered this:

```
    let uri = "http://[]@[";
    let uri = uri.parse::<http::Uri>().unwrap(); // is ok
    uri.host(); // panics
```

The panic is `parsing should validate brackets` and come from here: https://github.com/hyperium/http/blob/v0.2.1/src/uri/authority.rs#L486

I expect this to surface as an `Err` and panic on `.unwrap()`.

aeryz:
Fixes #435

### DIFF
--- a/src/uri/authority.rs
+++ b/src/uri/authority.rs
@@ -644,4 +644,10 @@ mod tests {
             .unwrap_err();
         assert_eq!(err.0, ErrorKind::InvalidUriChar);
     }
+
+    #[test]
+    fn rejects_invalid_use_of_brackets() {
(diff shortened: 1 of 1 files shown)


seanmonstar (commented):
This looks good to me, it'd be good to add the unit tests the value that was triggering the panic.

aeryz:
I added a test which tests the exact string that causes the bug.

seanmonstar (approved):
Thanks!

algesten:
Thanks!

PR #447, opened 2020-11-25
add Uri#to_builder() to easily modify a part of Uri

gfx:
It is messy to modify a `Uri` instance with `Uri::Builder`. For example, if one wants to modify `path_and_query`, they should do something like this:

```rust
let uri: &Uri = req.uri();
let path_and_query: String = "...";
let parts = uri.clone().into_parts();
let modified_uri = Uri::builder()
    .scheme(parts.scheme.unwrap())
    .authority(parts.authority.unwrap())
    .path_and_query(path_and_query.as_str())
    .build()
    .unwrap();
```

With this PR, it will become:

```rust
let uri: &Uri = req.uri();
let path_and_query: String = "...";
let modified_uri = uri.to_builder()
    .path_and_query(path_and_query.as_str())
    .build()
    .unwrap();
```

It will let us focus on the concern.

What do you think of it?

### DIFF
--- a/src/uri/mod.rs
+++ b/src/uri/mod.rs
@@ -673,6 +673,21 @@ impl Uri {
     fn has_path(&self) -> bool {
         !self.path_and_query.data.is_empty() || !self.scheme.inner.is_none()
     }
+
+    /// make a Uri::Builder from the Uri to modify part of it
+    pub fn to_builder(&self) -> Builder {
+        let mut builder = Self::builder();
+        if let Some(scheme) = self.scheme() {
+            builder = builder.scheme(scheme);
+        }
+        if let Some(authority) = self.authority() {
+            builder = builder.authority(authority);
+        }
+        if let Some(path_and_query) = self.path_and_query() {
+            builder = builder.path_and_query(path_and_query);
+        }
+        builder
+    }
 }
 
 impl<'a> TryFrom<&'a [u8]> for Uri {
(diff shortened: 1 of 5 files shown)


djc:
Unofficial review: this looks good to me; I think it's solving a real problem with a nicely compact API.

robjtede:
Alternative design:  
`Builder::from_parts()` / (`impl From<Parts> for Builder`)
construction becomes `Builder::from_parts(uri.to_parts())` / (`Builder::from(uri.to_parts())`).

Pros:
- equally comprehensible
- mirrors `Uri::from_parts` method
- (permits `parts.into()` where Builder is passed to fn)  

Cons:
- uglier when chained
- requires `Builder` import

src/uri/mod.rs
```diff
@@ -673,6 +673,22 @@ impl Uri {
     fn has_path(&self) -> bool {
         !self.path_and_query.data.is_empty() || !self.scheme.inner.is_none()
     }
+
+    /// make a Uri::Builder from the Uri to modify part of it
+    pub fn to_builder(self) -> Builder {
```
robjtede, line 678: naming nitpick, this would conventionally be `into_builder`

gfx: Thanks. It makes sense. Fixed at 1db81950d55a642dabd937fe2acf012e22eebfd8

reu:
Came looking for this! Any chance we can get this reviewed by a maintainer?

PR #448, opened 2020-11-30
Non-failing Uri Builder.

kaj:
Fixes one part of #330.  This makes it possible to create an `Uri` using the builder pattern without hanlding errors.

This is really a breaking change to `uri::Builder`, but rather than suggesting a breaking change directly, I introduce an alternative `uri::Builder2`. If this is accepted, the original Builder is deprecated, and in a later release Builder2 can be renamed back to Builder.

The new `Builder2` is generic on its contents, which starts out as the empty type `()`.  The parts `Scheme`, `Authority` and `PathAndQuery` can be added in any combination, and the `build` method is implemented for valid combinations only.

### DIFF
--- a/src/uri/builder.rs
+++ b/src/uri/builder.rs
@@ -165,6 +165,7 @@ mod tests {
 
     #[test]
     fn build_from_str() {
+        #[allow(deprecated)]
         let uri = Builder::new()
             .scheme(Scheme::HTTP)
             .authority("hyper.rs")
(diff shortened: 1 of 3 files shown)


kaj:
Looking at the generated documentation for `uri::Builder2`, it gets a bit repetitive, as their is a _property_ and a _try_property_ for each of the three properties is repeated for each of the eight variants of the `Builder2`, and a `build` method for four of them.  Does anyone have a suggestion for improving that?

dekellum:
Regarding the repetitive docs, is there some way to improve it via moving common methods to (still specific) trait implementations? Rustdoc currently hides those pretty well, though, perhaps more than would help.

kaj:
Replaced by #483 .

PR #454, opened 2020-12-13
Implement a non-failing Request builder.

kaj:
Fixes one part of #330. Creating a `Request` with the builder pattern should not require handling errors unless opting in by using a `try_`-method (which returns a `Result<Builder2>`, not a `Builder2` with an internal error state).

This is really a breaking change to `request::Builder`, but rather than suggesting a breaking change directly, I introduce an alternative `response::Builder2`. If this is accepted, the original `response::Builder` is deprecated, and in a later release `Builder2` can be renamed back to `Builder`.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -9,26 +9,28 @@
 //!
 //! Creating a `Request` to send
 //!
-//! ```no_run
-//! use http::{Request, Response};
+//! ```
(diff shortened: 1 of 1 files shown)


kaj:
Replaced by #483 .

PR #464, opened 2021-01-22
Added benchmarks for parsing method

a1ien:
I tried to optimize parsing 'Method' and failed ;). But at least there will be a benchmark

### DIFF
--- a/benches/method.rs
+++ b/benches/method.rs
@@ -1,0 +1,40 @@
+#![feature(test)]
+
+extern crate test;
+
+use http::method::Method;
+use test::Bencher;
+
+fn make_all_methods() -> Vec<Vec<u8>> {
+    vec![
+        b"OPTIONS".to_vec(),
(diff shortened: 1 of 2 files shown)


benches/method.rs
```diff
@@ -0,0 +1,40 @@
...
+        b"CONNECT".to_vec(),
+        b"PATCH".to_vec(),
+        b"CUSTOM_SHORT".to_vec(),
+        b"CUSTOM_LONG_METHOD".to_vec(),
+    ]
+}
+
+#[bench]
+fn method_easy(b: &mut Bencher) {
+    let name = b"Get";
```
seanmonstar, line 26: Should this be `GET`? With lowercase, it will parse as a "custom".

a1ien: yehh GET :) 
Fix

seanmonstar (approved):

PR #466, opened 2021-02-27
Make StatusCode::as_str return &'static str

RAnders00:
The lifetime of the returned `&str` was implicitly `pub fn as_str<'a>(&'a self) -> &'a str`. This changes that to `pub fn as_str<'a>(&'a self) -> &'static str`. This would allow me to use the returned `&str` in my project without having to clone it into an owned `String` first.

### DIFF
--- a/src/status.rs
+++ b/src/status.rs
@@ -132,7 +132,7 @@ impl StatusCode {
     /// assert_eq!(status.as_str(), "200");
     /// ```
     #[inline]
-    pub fn as_str(&self) -> &str {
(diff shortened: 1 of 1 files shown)


dekellum:
Dupe of #452 . Any of these can be merged, FFS. Had I known this would be so hard to land I would have snuck it into a larger PR.

seanmonstar:
Yea, this is a dupe. For reference, I've just started an issue (#468) to try to collect thoughts on the decision to make regarding this method.

dekellum:
Dup of #452 (for the record).

PR #470, opened 2021-03-09
Bring back `TryFrom<Bytes>` implementations

### LINKED ISSUE #459: Expose Bytes to the public API again
`http` is using the stable version of `bytes` crate now (#457) so I guess #369 can be reverted, except for the addition of `from_maybe_shared` constructors until `http` v0.3 or v1.0.

tesaguri:
This is mostly a revert of commit 4ce5e6a3a33c6548f885972db6f23450be18e133, but this does not bring back the error variants suffixed with `Bytes` (db9b1b9a76ce9c0b366fc6fff433ca4ca00571d1).

This also replaces usages of the internal `from_shared` associated functions with `try_from`.

Closes #459.

### DIFF
--- a/benches/header_value.rs
+++ b/benches/header_value.rs
@@ -2,6 +2,8 @@
 
 extern crate test;
 
+use std::convert::TryFrom;
+
 use bytes::Bytes;
 use http::HeaderValue;
 use test::Bencher;
@@ -14,7 +16,7 @@ fn from_shared_short(b: &mut Bencher) {
     b.bytes = SHORT.len() as u64;
     let bytes = Bytes::from_static(SHORT);
     b.iter(|| {
-        HeaderValue::from_maybe_shared(bytes.clone()).unwrap();
+        HeaderValue::try_from(bytes.clone()).unwrap();
     });
 }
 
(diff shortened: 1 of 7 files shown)


tesaguri:
Rebased onto master and updated the commit to retain `from_shared` functions for now to make the diff concise and easy to review.

tesaguri:
Rebased onto master and replaced the internal `from_shared` constructors with `try_from` in a separate commit.

Also, I've added a commit to remove the `from_maybe_shared` constructors because Hyper is at the release candidate stage now and these constructors do not seem to be intended for v1.0.

tesaguri:
@seanmonstar Sorry for the ping, but I believe this needs a decision, be it *for* or *against*, before the v1.0 release, since the PR contains a breaking change.

seanmonstar:
What's the breakage? The removal of `from_maybe_shared`? If done after 1.0, that could just be hidden and deprecated, it's not bad, right?

tesaguri:
All right. Then, I'll revert the removal of `from_maybe_shared` if/when the crate hits v1.0 before merging the PR.

tesaguri:
Removed the commit to remove the `from_maybe_shared` constructors, and added a commit to deprecate and hide them from docs instead.

The CI failure is of the Nightly one and I haven't touched anything related to the `Sealed` trait, so it's not relevant to the patch, I guess?

PR #474, opened 2021-04-02
Allow `{`, `"` and `}` in URI paths

nox:


seanmonstar (approved):
> send requests with JSON directly embedded in the URI path. Yes, those things happen for real.

😱

PR #477, opened 2021-04-08
Add missing TryFrom<String> and TryFrom<Vec<u8>> impls.

Arnavion:
Also fix PathAndQuery's TryFrom<String> impl to not copy the bytes into
a new Bytes.


Arnavion:
@seanmonstar Ping.

seanmonstar (approved):

seanmonstar:
Thanks for catching this!

PR #478, opened 2021-04-09
Initial fuzzer set up.

DavidKorczynski:
Hi! 

I was working on fuzzing hyperium-http, and thought it would be nice to integrate it with OSS-Fuzz (https://github.com/google/oss-fuzz/pull/5330). I think it would be nice to have the entire set of hyperium repositories in under one umbrella OSS-Fuzz project, so let me know if you are interested in this and I will set it all up.

### DIFF
--- a/fuzz/src/fuzz_http.rs
+++ b/fuzz/src/fuzz_http.rs
@@ -1,0 +1,28 @@
+#![no_main]
+
+use libfuzzer_sys::fuzz_target;
+use libfuzzer_sys::arbitrary::Arbitrary;
+use http::Request;
+use http::StatusCode;
+use http::Response;
+
+#[derive(Debug, Arbitrary)]
+struct HttpSpec {
+        uri: Vec<u8>,
+        header_name: Vec<u8>,
+        header_value: Vec<u8>,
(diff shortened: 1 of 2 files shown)


fuzz/src/fuzz_http.rs
```diff
@@ -0,0 +1,28 @@
...
+use http::StatusCode;
+use http::Response;
+
+#[derive(Debug, Arbitrary)]
+struct HttpSpec {
+        uri: Vec<u8>,
+        header_name: Vec<u8>,
+        header_value: Vec<u8>,
+        status_codes: Vec<u8>,
+        response: Vec<u8>,
```
seanmonstar, line 15: Looks like this is just setting a `Vec` as the response body, which allows any type and any value. So I think generating this vector won't be useful for the fuzzing.

seanmonstar (commented):
I think this is a great idea, thanks!

fuzz/src/fuzz_http.rs
DavidKorczynski: Thank you! This should be solved now.

seanmonstar (approved):

PR #479, opened 2021-04-16
Add TryFrom implementations for owned types for HeaderName

sagebind:
`HeaderValue` implements `TryFrom` over both reference and owned types, but `HeaderName` does not. This seemed like an accidental omission to me rather than something intentionally absent. Including this conversion slightly increases ergonomics by allowing you to create a `HeaderMap` from a `HashMap<String, String>` and in similar other constraints.

See also https://github.com/sagebind/isahc/issues/314.


seanmonstar (approved):
Yep, looks like an oversight to me, thanks!

PR #490, opened 2021-07-13
Add author copyrights to the header map file

### LINKED ISSUE #489: Missing author information/attribution
When I read the HeaderMap code I recognize that it's based on my own code in crate indexmap (at one time formerly "ordermap"). The license is Apache-2 or MIT but it would be nice - if this is true - if authorship information was retained, as it has to be.

HeaderMap source link https://github.com/hyperium/http/blob/e54da7175a9af490a2e04003c5b49ff259222a88/src/header/map.rs

It seems to come all the way from the initial implementation https://github.com/hyperium/http/pull/13

When I read the code I recognize the probe loop, insert split into mulitple phases, the struct Pos, type HashValue etc. We have since refactored and upgraded IndexMap a lot, so they have certainly diverged.

An older version of indexmap source (2016) is here: https://github.com/bluss/indexmap/blob/02a9bf73356eb976ddcb980d1b839db6c2ee122f/src/lib.rs  Comparing for example the function `do_insert_phase_two` in HeaderMap, with `insert_phase_2` in indexmap, I'd say the former is derived from the latter.

If you credit me, please write Ulrik Sverdrup "bluss" as the author. Thank you!

bluss:
The HeaderMap implementation as a whole is mostly orignal to http-rs,
but the implementation was based on and derived from the indexmap crate
(formerly ordermap) from circa 2017, from which the structure and many
methods were adapted. For this reason, mention the two indexmap authors
(one of them is me) from that time.

Fixes #489

### DIFF
--- a/src/header/map.rs
+++ b/src/header/map.rs
@@ -1,3 +1,12 @@
+// Copyright 2017-2021 http-rs authors
+// Copyright 2016-2017 Ulrik Sverdrup "bluss", Piotr Czarnecki
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
 use std::collections::HashMap;
 use std::collections::hash_map::RandomState;
 use std::convert::TryFrom;


bluss:
Any other solution is fine to me too, just making a draft.

src/header/map.rs
```diff
@@ -1,3 +1,12 @@
+// Copyright 2017-2021 http-rs authors
```
seanmonstar, line 1: I think "http-rs" is actually a separate organization. Maybe we just leave off about the rest of the copyright, assuming the top level license works, and we just say "Parts of this file are ..."?

bluss: Ok. http-rs authors is exactly what the top-level license files write too.

PR #496, opened 2021-08-09
feature: make possible to set port number in Uri::builder

Dushistov:


src/uri/tests.rs
```diff
@@ -517,3 +517,37 @@ fn test_partial_eq_path_with_terminating_questionmark() {
...
+            .build()
+            .unwrap()
+    );
+    Uri::builder()
+        .scheme("ws")
+        .authority("localhost:8000")
+        .port(8000u16)
+        .path_and_query("/demo/")
+        .build()
+        .unwrap_err();
```
robjtede, line 552: Can you elaborate on why this shouldn't work. I guess it's because order would matter but want to see that documented if so.

Dushistov: The idea that if port is set via "localhost:8000", then subsequent setting of port generate error

robjtede: What I mean is that I can see it being useful to override the port of an authority string you've acquired from elsewhere.

PR #499, opened 2021-08-20
Make HeaderName::from_static const

jeddenlea:
... plus some clean-up.

It was only after I came up with the scheme using
`const fn from_bytes(&[u8]) -> Option<StandardHeader>`
that I noticed the debug+wasm32-wasi version of `parse_hdr`, which had
something very similar.

While cleaning up that function, I realized it still would still panic
if an attempted name was too long, which had been fixed for all other
targets and profiles in #433.  Then, I thought it would be worth seeing
if the use of `eq!` in the primary version of `parse_hdr` still made any
difference.

And, it would not appear so. At least not on x86_64, nor wasm32-wasi run
via wasmtime.  I've run the benchmarks a number of times now, and it
seems the only significant performance change anywhere is actually that
of `HeaderName::from_static` itself, which now seems to run in about 2/3
the time on average.

Unfortunately, `const fn` still cannot `panic!`, but I've followed the
lead from `HeaderValue::from_static`.  While that version required 1.46,
this new function requires 1.49.  That is almost 8 months old, so
hopefully this isn't too controversial!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
           - nightly
           # When updating this value, don't forget to also adjust the
           # `rust-version` field in the `Cargo.toml` file.
-          - 1.46.0
+          - 1.49.0
 
         include:
           - rust: nightly
--- a/src/header/name.rs
+++ b/src/header/name.rs
@@ -64,7 +64,7 @@ macro_rules! standard_headers {
     (
         $(
             $(#[$docs:meta])*
-            ($konst:ident, $upcase:ident, $name:expr);
+            ($konst:ident, $upcase:ident, $name_bytes:literal);
         )+
     ) => {
         #[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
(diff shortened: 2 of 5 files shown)


seanmonstar:
@carllerche you wrote the `eq!` stuff, any insight on instances where it's worth keeping?

carllerche:
It's been a while. My recollection is this was to optimize header parsing. If there are benchmarks still checking parse speed and they aren't impacted, then it is probably fine.

jeddenlea:
Using `cargo +nightly bench header_name`, before:

```
test header_name_bad     ... bench:           9 ns/iter (+/- 0)
test header_name_easy    ... bench:          10 ns/iter (+/- 0)
test header_name_various ... bench:       3,885 ns/iter (+/- 875)
```

and with this change,

```
test header_name_bad         ... bench:          11 ns/iter (+/- 1)
test header_name_custom      ... bench:          81 ns/iter (+/- 5)
test header_name_easy        ... bench:          11 ns/iter (+/- 0)
test header_name_from_static ... bench:       1,024 ns/iter (+/- 483)
test header_name_various     ... bench:       3,829 ns/iter (+/- 288)
```

jeddenlea:
.... and just to be complete, here are the two new benchmarks cut out and run on 0.2.6,

```
test header_name_custom      ... bench:          79 ns/iter (+/- 7)
test header_name_from_static ... bench:       1,241 ns/iter (+/- 72)
```

So, maybe only a 20% improvement instead of a 33% improvement, but still!

npmccallum:
@jeddenlea @seanmonstar FYI - const_panic is now stable in `1.57.0`.

src/header/name.rs
```diff
@@ -1760,33 +1218,31 @@ impl HeaderName {
...
-                },
-                Repr::Custom(MaybeLower { buf: _, lower: false }) => {
-                    // With lower false, the string is left unchecked by
-                    // parse_hdr and must be validated manually.
-                    for &b in bytes.iter() {
-                        if HEADER_CHARS_H2[b as usize] == 0 {
-                            panic!("invalid header name")
-                        }
-                    }
+    #[allow(unconditional_panic)] // required for the panic circumventon
```
jeddenlea, line 1221: D'oh, I copied the same error fixed in #532

src/header/name.rs
```diff
@@ -1727,12 +1163,34 @@ impl HeaderName {
...
+    ///
+    /// ```text
+    /// error: any use of this value will cause an error
+    ///     --> http/src/header/name.rs:1241:13
+    ///      |
+    /// 1241 |             ([] as [u8; 0])[0]; // Invalid header name
+    ///      |             ^^^^^^^^^^^^^^^^^^
+    ///      |             |
+    ///      |             index out of bounds: the length is 0 but the index is 0
+    ///      |             inside `fastly::http::HeaderName::from_static` at http/src/header/name.rs:1241:13
```
LucioFranco, line 1186: did you mean to include `fastly` here?

LucioFranco (approved):
overall lgtm w/ the benches you posted.

src/header/name.rs
seanmonstar:
```suggestion
    ///      |             inside `http::HeaderName::from_static` at http/src/header/name.rs:1241:13
```

seanmonstar (approved):
With the one proposed change, we can merge this :)

src/header/name.rs
jeddenlea: Haha, no, I did not.  I'm so used to seeing it I didn't even notice.  I'm going to force push that change.

seanmonstar (approved):

.github/workflows/ci.yml
```diff
@@ -21,7 +21,7 @@ jobs:
           - nightly
           # When updating this value, don't forget to also adjust the
           # `rust-version` field in the `Cargo.toml` file.
-          - 1.46.0
+          - 1.49.0
```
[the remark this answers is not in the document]
LucioFranco: This project follows the msrv of tokio https://github.com/tokio-rs/tokio/#supported-rust-versions but we should add it to the changelog. Good catch.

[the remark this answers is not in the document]
LucioFranco: I think this is mainly due to a few reasons, tokio is already 1.0 and thus at the time of 1.0 we try to support everyone. While this crate is below 1.0 so the guarantees are different. I assume it should be possible to pin the version of http though for your crate?

PR #500, opened 2021-08-23
optionally support ahash

glebpom:
aHash is known to be one of the most effective hash implemented on rust [see link](https://github.com/tkaitchuck/aHash/blob/master/compare/readme.md). 

This PR introduces the optional usage of aHash instead of fnv hash. 

In my benchmark of HTTP server (hyper/warp) it shows ~1% in performance improvement overall. 

Alternatively we can fully migrate to aHash without introducing the feature, because it's the internal implementation details.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,8 +23,9 @@ edition = "2018"
 
 [dependencies]
 bytes = "1"
-fnv = "1.0.5"
 itoa = "0.4.1"
+fnv = { version = "1.0.5", optional = true }
+ahash = { version = "0.7", optional = true }
 
 [dev-dependencies]
 indexmap = "1.0"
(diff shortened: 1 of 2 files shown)


vilgotf:
fyi: this is a breaking change since `default-features = no` becomes a compile time error

dekellum:
Any more specific benchmark results local to http? 

> this is a breaking change since default-features = no becomes a compile time error

Can we avoid that by doing something like using fnv when `#[cfg(not(feature = "ahash"))]` and possibly making the fnv dependency non-optional but a no-op in the ahash case?  (An unused dep for some seems better then breaking builds.)  Related, adjust CI to all test your ahash feature?

glebpom:
Actually, I didn't find the real benefits of this change with the further benchmarks. Even more, in some cases, it worked slower.

hawkw:
For the record, my (admittedly limited) understanding is that there are [some concerns about the hashDoS resistance of aHash](https://internals.rust-lang.org/t/a-new-default-hasher-for-hashmap/9747/19). Since `http::HeaderMap`s are, naturally, exposed to untrusted input from the network, we should probably be cautious about replacing the hashing algorithm with a non-cryptographically secure hash with dubious DoS-resistance claims...

PR #503, opened 2021-09-08
Add Cross-Origin-{Embedder/Resource}-Policy Headers

theduke:
Adds two new typed headers.


seanmonstar:
Thanks for the PR!

The list of constants in `http` is mostly for the _most_ common headers that a user would normally reach for. These new headers seem quite niche, so I'd probably suggest that they don't need to be in `http` and someone can just make a constant in their own code.

migueldemoura:
Hey @seanmonstar, could we reconsider this? These headers are used far more widely now, and it would be nice to have them typed as we do for the other common ones.

PR #505, opened 2021-09-20
Add optional borsh serialization for request/response

XX:


seanmonstar:
Thanks for the PR! But I don't believe we'll have direct support most crates in `http`. The more public dependencies there are, the more fragile it becomes. If one of the dependencies needs to make a breaking change, then `http` is left with an annoying choice for users: either upgrade as well and force a breaking change, or stay behind but stop working with the new version.

PR #510, opened 2021-11-08
status: make as_str return &'static str

jhgg:
This is technically a 'static str, however we return it as not that. 

Is there a reason, can we make this a 'static str? This will help resolve some lifetime around metrics we are encountering.


CryZe:
Duplicate of #452

jhgg:
Aaaa, would love to see #452 merged, will close this PR!

PR #511, opened 2021-11-10
Add `Request::parts()` and `Request::parts_mut()`

olix0r:
`http::Request::into_parts()` is the only way to access a request's
`Parts` structure.

It can be convenient to write utility code that takes a `&Parts` or
`&mut Parts`, i.e., so there's no need to parameterize the utility on
body type if it can't be used.

This change introduces accessors for the `Parts` structure so callers
need not use `Request::into_parts` and `Request::from_parts` to access
the `Parts` structure.

---

In service of https://github.com/tower-rs/tower-http/issues/154

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -53,7 +53,7 @@
 //! ```
 
 use std::any::Any;
-use std::convert::{TryFrom};
+use std::convert::TryFrom;
 use std::fmt;
 
 use crate::header::{HeaderMap, HeaderName, HeaderValue};
@@ -231,7 +231,6 @@ impl Request<()> {
     where
         Uri: TryFrom<T>,
         <Uri as TryFrom<T>>::Error: Into<crate::Error>,
-
     {
         Builder::new().method(Method::GET).uri(uri)
     }
@@ -254,7 +253,6 @@ impl Request<()> {
     where
         Uri: TryFrom<T>,
         <Uri as TryFrom<T>>::Error: Into<crate::Error>,
-
     {
         Builder::new().method(Method::PUT).uri(uri)
     }
@@ -277,7 +275,6 @@ impl Request<()> {
     where
         Uri: TryFrom<T>,
         <Uri as TryFrom<T>>::Error: Into<crate::Error>,
-
     {
         Builder::new().method(Method::POST).uri(uri)
     }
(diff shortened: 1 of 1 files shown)


olix0r:
It looks like my editor ran rustfmt which modified some whitespace. I can back out those changes if desired.

olix0r:
@seanmonstar @carllerche I'm curious if there objections to adding these accessors.

cc @davidpdrsn

seanmonstar:
It's worth pointing out that by adding an accessor like this, we are then stuck with keeping the internal shape the same forever, since it's essentially opening up as public fields. This differs from `into_parts`, which could move from a changed structure into the `Parts` shape.

olix0r:
@seanmonstar that's a fair point.

in favor of this: it would allow a function that takes `&http::Request<_>` to access the `&Parts`.

Maybe an alternative would be to add a `PartsRef<'_>` that holds only a reference the headers/extensions/uri? This seems awkward for other reasons (we would probably have to drop the `_mut` variant).

I don't feel strongly about this, but it's something I've run into more than once; so I wanted to start a discussion.

olix0r:
Closing for now, as this will probably ossify the API more than we want. Happy to reopen in the future, though.

jplatte:
Can this be re-opened? Of course it's a tradeoff, but it doesn't seem like anybody was strongly against committing to `Request` holding `request::Parts` / `Response` holding `response::Parts`. It would simplify the implementation of `RequestExt` that we just added in axum, and also future-proof it for any additional `Parts` request / response bits.

seanmonstar:
Ive been recently thinking we could make the size of `Request` smaller if we possibly boxed some fields. I still don't see a reason to commit to holding a `Parts` internally. It's the same as just making all the fields public.

jplatte:
Well, the reason would be that it's useful for downstream projects, as shown twice now. Maybe a solution could be to (in the next breaking release) have the `Parts` types be as opaque as `Request` / `Response` in terms of fields, and provide accessors as well? Then a change like boxing fields could be done across all types at the same time and `parts()` and `parts_mut()` could exist too.

EFanZh:
Hi @seanmonstar, How about an alternative design like this:

```rust
pub struct PartsMut<'a, T> {
    pub status: &'a mut StatusCode,
    pub version: &'a mut Version,
    pub headers: &'a mut HeaderMap<HeaderValue>,
    pub extensions: &'a mut Extensions,
    pub body: &'a mut T,
    _priv: (),
}

impl<T> Request<T> {
    // The name can be
    pub fn parts_mut(&mut self) -> PartsMut<'_, T> {
        PartsMut {
            ...
        }
    }
}
```

In this way, we can allow user to hold mutable references to different parts without exposing the concrete implementation.

PR #512, opened 2021-11-11
README: Remove obsolete `extern crate http` instructions

Turbo87:
This crate declares `edition = 2018`, in which these `extern crate` instructions are no longer necessary, so it should be safe to remove them from the README file.


seanmonstar (approved):
Thanks, nice catch!

PR #515, opened 2021-12-01
ensure URIs with authority have leading slash on path

### LINKED ISSUE #507: path_and_query allow path not starting with slash
using [`path_and_query()`](https://docs.rs/http/0.2.5/http/uri/struct.Builder.html#method.path_and_query):

 ```rust
Uri::builder()
    .scheme(Scheme::HTTP)
    .authority("localhost:8000")
    .path_and_query("hello/world")
    .build()
```

report no error, but rocket didn't accept it with:

```none
Oct 12 19:24:35.028 DEBUG hyper::proto::h1::conn: parse error (invalid URI) with 115 bytes
Oct 12 19:24:35.028 DEBUG hyper::proto::h1::role: sending automatic response (400 Bad Request) for parse error
```

I believe this should be considered a bug either http or rocket, I needed to add a slash at the beginning `"/hello/world"`

robjtede:
partially addresses #507

### DIFF
--- a/src/byte_str.rs
+++ b/src/byte_str.rs
@@ -20,7 +20,7 @@ impl ByteStr {
     #[inline]
     pub fn from_static(val: &'static str) -> ByteStr {
         ByteStr {
-            // Invariant: val is a str so contains vaid UTF-8.
+            // Invariant: val is a str so contains valid UTF-8.
             bytes: Bytes::from_static(val.as_bytes()),
         }
     }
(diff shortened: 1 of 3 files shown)


kolbma:
I think this would only deny to create wrong URIs on client side, but server side pushes invalid URIs with HTTP/2 to the service handler.  
How do we create invalid URIs if the Builder checks this?

robjtede:
I've removed the link to #507 since it doesn't fully address it. But it does address the builder side of the problem afaict.

PR #521, opened 2022-01-25
Enable simpler header name parser in debug builds

ibraheemdev:
This shaves close to a second off debug builds on my machine.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,12 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
+      
+      - name: Test Release
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --release
 
       - name: Test all benches
         if: matrix.benches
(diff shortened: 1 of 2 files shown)


hawkw:
If we're going to do this, it seems like it might also be a good idea to add a CI step to run the tests in release mode? Otherwsie, most of the parser code is no longer tested on CI.

seanmonstar:
Maybe I'm misunderstanding... Seems this is turning off the debug assertions?

ibraheemdev:
It makes it so the simpler version of `parse_hdr` in any debug builds, instead of only in WASM debug builds. `debug_assertions` is the cfg flag that identifies a debug build (ish).

seanmonstar:
I'd normally think in a debug build we'd want the  extra checks.

jeddenlea:
https://github.com/hyperium/http/pull/499 removes the `eq` macro in all cases.

ibraheemdev:
Closing in favor of #499.

PR #522, opened 2022-02-05
moved to finish_non_exhaustive in Debug impl

06chaynes:
This moves from the `finish` method to the `finish_non_exhaustive` method in Debug impl blocks where applicable.

"Marks the struct as non-exhaustive, indicating to the reader that there are some other fields that are not shown in the debug representation." - [std::fmt::DebugStruct Docs](https://doc.rust-lang.org/std/fmt/struct.DebugStruct.html#method.finish_non_exhaustive)


olix0r:
Note that this method is only available since 1.53.0, so this change impacts the MSRV of `http`

PR #524, opened 2022-02-17
HeaderValue::from_str validate if string is visible ascii

### LINKED ISSUE #519: HeaderValue::from_str does not validate that the string is valid ascii
The documentation of `HeaderValue::from_str` says that:

https://github.com/hyperium/http/blob/abe651273f4cf19cf9a247f376e9ece85becc722/src/header/value.rs#L101-L104

So I would expect this function to return an error when I pass it a non-ascii utf8 string which contains byte values in the 128-255 range. However the result is success.

[Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ea874d2d8d98558307a7d983640e7d4c)

```rust
let x = http::header::HeaderValue::from_str("ñ");
panic!("{:?}", x);
```

```
thread 'main' panicked at 'Ok("\xc3\xb1")', src/main.rs:3:5
```

xoac:
close #519

The documentation say that valid ASCII is between (32..127) char. But function `is_visible_ascii()` check also for `\t` char what is coded as `9`.

### DIFF
--- a/src/header/value.rs
+++ b/src/header/value.rs
@@ -123,7 +123,15 @@ impl HeaderValue {
     /// ```
     #[inline]
     pub fn from_str(src: &str) -> Result<HeaderValue, InvalidHeaderValue> {
-        HeaderValue::try_from_generic(src, |s| Bytes::copy_from_slice(s.as_bytes()))
+        for &b in src.as_bytes() {
+            if !is_visible_ascii(b) {
+                return Err(InvalidHeaderValue { _priv: () });
+            }
+        }
+        Ok(HeaderValue {
+            inner: Bytes::copy_from_slice(src.as_bytes()),
+            is_sensitive: false,
+        })
     }
 
     /// Converts a HeaderName into a HeaderValue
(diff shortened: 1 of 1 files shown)


lilyball:
Your PR description doesn't match the code. `is_visible_ascii()` actually allows for `\t` (even though it's not visible; I think the name here is wrong, it should be called `is_valid_ascii` instead). And what your PR does is restrict `from_str` to ASCII, which has nothing to do with `\t`.

xoac:
I just matched implementation of [`from_static()`](https://github.com/hyperium/http/blob/c28945c6c6f99379b674a1e961a743c7752f2346/src/header/value.rs#L83) that has exactly the same description for ASCII requirements as `from_str` and use the same function (`is_visible_ascii`) to check correctness but panic instead of returning error. 

So the code should be correct if `from_static()` is correct.

lilyball:
Note: I put a number of comments in the linked issue (https://github.com/hyperium/http/issues/519).

PR #526, opened 2022-02-22
Use HeaderValue::from_static instead of parsing and unwrapping in HeaderMap example

Herbstein:
This change is in response to hyperium/hyper#2714.


src/header/map.rs
```diff
@@ -29,8 +29,8 @@ pub use self::into_header_name::IntoHeaderName;
 /// # use http::header::{CONTENT_LENGTH, HOST, LOCATION};
```
mnpw, line 29:
```suggestion
/// # use http::HeaderValue;
/// # use http::header::{CONTENT_LENGTH, HOST, LOCATION};
```

`HeaderValue` import is required to use `HeaderValue::from_static`. Rest looks good!

mnpw (changes requested):
Import addition requested, rest looks good.

LucioFranco:
I think this needs a rebase and then we can likely merge but we need to get CI running first.

Herbstein:
Hi. I completely missed the comments on this because of the delay between posting and commenting. I've updated the PR with the suggested changes and rebased on the latest master.

robjtede (approved):

PR #528, opened 2022-03-01
Allow percent encoding in host

paullgdc:
According to RFC 3986 https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2 , the host section of the authority of an url is allowed to carry percent-encoded characters.

The parsing code only allowed `%` in ip adresses, and userinfo, which is both needlessly restrictive, and allows some invalid urls if there isn't two hex characters after the percent character.

This PR allows percent encoding everywhere in the host, and checks if the percent-encoding is valid.

This is useful for instance to implements Unix domain sockets URI, that would look like `unix://<percent encoded socket path>/<http request path>`


src/uri/authority.rs
```diff
@@ -90,7 +90,7 @@ impl Authority {
                     colon_cnt += 1;
                 }
                 b'[' => {
-                    if has_percent || start_bracket {
+                    if start_bracket {
                         // Something other than the userinfo has a `%`, so reject it.
```
robjtede, line 94: comment is not relevant any more

robjtede:
Torn on whether this is a good thing for the crate, here's some facts though:
- rfc3986 defines a _generic_ URL syntax that allows percent encoding in the authority
- rfc7230 references rfc3986 section 3.2 without additional stipulations
- [`httparse` allows % in URIs](https://github.com/seanmonstar/httparse/blob/master/src/lib.rs#L58)
- the WHATWG URL standard seems to [disallow percent encoding in domain parts](https://url.spec.whatwg.org/#host-miscellaneous) but not in the authority as a whole
  - [relevant webplatform test](https://github.com/web-platform-tests/wpt/blob/master/url/resources/urltestdata.json#L3657)

Given these, and without requiring an up to date public suffix list referenced by the WHATWG standard, it feels like allowing percent encoding as this PR does would be largely standard compliant.

PR #533, opened 2022-03-18
Add Natvis visualizations for some `http` types

ridwanabdillahi:
This change adds Natvis visualizations for some `http` types to help improve the debugging experience on Windows.

Debugging types such as `Request` and/or `Response` under a Windows debugger, WinDbg or the Visual Studio debugger, it can be a bit difficult to dissect exactly what is data is contained in each type. The Rust compiler does have Natvis support for some types, but this is limited to some of the core libraries and not supported for external crates.

There is currently an [RFC](https://github.com/rust-lang/rfcs/pull/3191) out to add support for embedding debugging visualizations such as Natvis in a generated PDB/ELF.

Natvis is a framework that can be used to specify how types should be viewed under a supported debugger, such as the Windows debugger (WinDbg) and the Visual Studio debugger.


ridwanabdillahi:
Hello @LucioFranco do you have any time to review this PR? This change adds Natvis visualizations for some of the core `http` types to ease debugging these types on Windows debuggers along with documentation on how to do so.

PR #537, opened 2022-03-28
Fix Uri docs

jeddenlea:
The doc comments for `impl From<Uri> for Parts` actually described
`Uri::from_parts`.


jeddenlea:
D'oh, this and #538 are too close together and cannot be cleanly merged one after the other.  So, I'm adding this as a second commit to #538.

PR #538, opened 2022-03-28
From impls of PathAndQuery and Authority for Uri

jeddenlea:
A Uri may logically hold only an Authority or only a PathAndQuery.
Should an application want to make such Uris, it would be easier to
safely create them by first making just an Authority or just a
PathAndQuery, then cheaply convert them directly into a Uri.

### DIFF
--- a/src/uri/mod.rs
+++ b/src/uri/mod.rs
@@ -736,6 +736,28 @@ impl<'a> TryFrom<&'a Uri> for Uri {
     }
 }
 
+/// Convert an `Authority` into a `Uri`.
+impl From<Authority> for Uri {
+    fn from(authority: Authority) -> Self {
+        Self {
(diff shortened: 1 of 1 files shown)


robjtede (approved):
Seems reasonable to add these shortcuts. Both forms are constructable in other ways.

Might be good to add a couple tests anyway just so that any future internal changes don't accidentally allow these impls to construct invalid URIs.

seanmonstar (approved):

jeddenlea:
@seanmonstar #499 required 1.49, but it looks like the CI is still trying to build with 1.46.

seanmonstar:
I think it's because it's commits from before the updated MSRV. The list says 1.49 is required but not running.

jeddenlea:
That makes sense too... I've rebased.

PR #540, opened 2022-04-01
Merge #428 and #499

jeddenlea:
For your consideration.  I'm not sure what the ideal way to collapse these commits together would be to match the clean, ordered scheme this repo is otherwise doing.  But, if we were to apply #428 atop #499, I think it would look like this.

### DIFF
--- a/src/header/name.rs
+++ b/src/header/name.rs
@@ -64,7 +64,7 @@ macro_rules! standard_headers {
     (
         $(
             $(#[$docs:meta])*
-            ($konst:ident, $upcase:ident, $name:expr);
+            ($konst:ident, $upcase:ident, $name_bytes:literal);
         )+
     ) => {
         #[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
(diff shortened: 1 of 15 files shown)


LucioFranco:
Hi @jeddenlea I think due to the nature of those two PRs I think it probably makes sense to review them individually and have separate commits. So we should try to revive those again even though there will be some conflicts.

jeddenlea:
@LucioFranco certainly! I won't be upset if this PR is abandoned, I just wanted to work out what a merge would look like because it would be great for both to be included in the next release, and there are lots of conflicts.  This can be used as a reference if that's helpful.  I did update it to include the small tweak I had to make on #499 for final approval.

seanmonstar (approved):

src/header/name.rs
```diff
@@ -1760,33 +1235,31 @@ impl HeaderName {
...
-                    for &b in bytes.iter() {
-                        if HEADER_CHARS_H2[b as usize] == 0 {
-                            panic!("invalid header name")
-                        }
-                    }
+    #[allow(unconditional_panic)] // required for the panic circumvention
+    pub const fn from_static(src: &'static str) -> HeaderName {
+        let name_bytes = src.as_bytes();
+        if let Some(standard) = StandardHeader::from_bytes(name_bytes) {
+            return HeaderName{
```
seanmonstar, line 1242: I'd think rustfmt would catch the missing space... Not that I care, but makes me wonder 🤔

seanmonstar:
Looks like we got conflicts from merging first your other PR. I've read through the code and changes, everything is super clear, thank you! Want to try to rebase and we'll get a release out with this?

seanmonstar:
Well I tried resolving the conflicts with GitHub's help, and it failed in one tiny way. So I've now fixed it up locally, and ended up squashing it into a single commit since the diff was quite small now that #499 was merged before. The new PR is in #555.

seanmonstar:
Merged in #555.

PR #545, opened 2022-04-15
Added test coverage for the StatusCode is_xxxxx() functions

ivd-git:
Just a small contribution that adds test coverage for the is_xxxxx() functions from StatusCode.


seanmonstar (approved):
Cool, thanks for this :D

PR #546, opened 2022-04-19
Add Extensions::extend

### LINKED ISSUE #525: Allow merging two `Extensions` maps
Motivation: https://github.com/tokio-rs/axum/issues/770

Prior art: The anymap crates allows this through `map1.extend(map2.into_raw())` (AFAICT, haven't tested)

jplatte:
Closes #525

### DIFF
--- a/src/extensions.rs
+++ b/src/extensions.rs
@@ -188,6 +188,39 @@ impl Extensions {
             .as_ref()
             .map_or(0, |map| map.len())
     }
+
+    /// Extends `self` with another `Extensions`.
+    ///
+    /// If an instance of a specific type exists in both, the one in `self` is overwritten with the
+    /// one from `other`.
+    /// 
+    /// # Example
+    /// 
+    /// ```
+    /// # use http::Extensions;
+    /// let mut ext_a = Extensions::new();
+    /// ext_a.insert(8u8);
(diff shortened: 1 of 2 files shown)


src/extensions.rs
```diff
@@ -188,6 +188,35 @@ impl Extensions {
...
+    /// ext_a.extend(ext_b);
+    /// assert_eq!(ext_a.len(), 3);
+    /// assert_eq!(ext_a.get::<u8>(), Some(&4u8));
+    /// assert_eq!(ext_a.get::<u16>(), Some(&16u16));
+    /// assert_eq!(ext_a.get::<&'static str>().copied(), Some("hello"));
+    /// ```
+    pub fn extend(&mut self, other: Self) {
+        if let (Some(this), Some(other)) = (&mut self.map, other.map) {
+            this.extend(*other);
+        }
```
neoeinstein, line 218: Do we need to handle the case where `self.map` is `None`, but `other.map` is `Some(_)`? Perhaps with the following:
```suggestion
        } else if self.map.is_none() {
            *self.map = other.map;
        }
```

jplatte: Oh, of course 🤦🏼

LucioFranco (approved):

PR #561, opened 2022-06-30
update: export HeaderName from crate root

Velfi:
We already export `HeaderMap` and `HeaderValue` from the crate root, so why not `HeaderName`?


robjtede (approved):

LucioFranco (approved):
Seems fine by me, cc @seanmonstar

seanmonstar (approved):

PR #562, opened 2022-07-23
Implement COOP and COEP

daxpedda:
See the Mozilla documentation on [COOP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Opener-Policy) and [COEP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Embedder-Policy).

i noticed that some of the documentation was basically copy-pasted from the Mozilla documentation, so I did the same. Happy to use my own wording if that's undesired.


robjtede (approved):

daxpedda:
Rebased after https://github.com/hyperium/http/pull/579.

perry-birch:
Working on a PR in a dependent lib which could use this if it's close to approval.

Updates look correct to me.

Vagelis-Prokopiou:
Hello all.

I would also suggest adding the following headers too:

- Cross-Origin-Resource-Policy
- X-Permitted-Cross-Domain-Policies
- Clear-Site-Data

They are headers recommended by OWASP (see https://owasp.org/www-project-secure-headers/#configuration-proposal), and I think that they should exist in `http` as constants.

PR #569, opened 2022-10-28
Make StatusCode::as_str() return a &'static str explicitly

fasterthanlime:
It's already indexing into a `&'static str` constant, I'd like to use it as a `&'static str` from my code, the only thing preventing that is the type signature :)


LucioFranco (approved):
This seems fine to me, would like to @seanmonstar to take a look as well

tot0:
YES!

CleanCut:
👋🏻 Hi! I'm wondering if this would be okay to merge now. 😄

peddermaster2 (approved):

PR #572, opened 2022-11-08
Add Permissions-Policy header

paolobarbolini:
Add [`Permission-Policy`](https://www.w3.org/TR/permissions-policy-1/) to the list of headers


seanmonstar:
Thanks for submitting the PR! My initial feeling is that this header is likely too niche to include in the general list of constants. There are other header names that we've not added as well. But I realize that there isn't policy spelled out for how to decide what names to add.

seanmonstar:
I filed #573 to track creation of that policy.

Issue #395, opened 2020-02-07
Update `extensions` to be cloneable


LucioFranco:
I propose that we move forward to allow `Request<B: Clone>` to `impl Clone`. The big blocker here is that `extensions` is not clonable. This could be fixed by making a AnyClone trait:

```rust
pub trait AnyClone: Any + Clone {}
```

The downside here is that all things you store in extensions is cloneable but I think this is a fair trade off since mostly its just information that could be thrown behind an arc or bytes.

cc @seanmonstar @carllerche @hawkw @davidbarsky

dekellum:
Some thoughts from an interested user:

Given new requirement on extensions, a breaking change then?

You would apply the same to `Response`?

It sounds like you would `impl Clone where B: Clone` so its kind of optional for bodies, but over time there might be more code expecting this.  An aspect that gives me pause is that hyper's default `Body` implements `Stream` (and is not `Clone` yet) and `http_body::Body` defines a `Stream`-like extension interface.  How should implementations handle cloning a `Stream`-type after it has been polled? Do cloned `Stream`s start from the beginning (if that's even possible) or from the point they are cloned?  

With synchronous corollaries, we would typically not clone an `Iterator`, but clone the thing that can provide a new iterator, e.g. `Vec` or any `impl IntoIterator`. In other words, adding a `Clone` requirement suggests that maybe, in a breaking release, we should also change to requiring a `Body` implement `IntoStream`?  (Note unfortunately there is already something called `IntoStream`, for an entirely different purpose, in futures 0.3.)

With clone in play, the `HeaderMap`s in `Request` and `Response` might benefit from `Arc` or `Cow`? 

Would the next breaking release also be time to consider integrating `http_body` into this crate? Or is that considered out of scope?  (As an observer I've hoped the second crate was a temporary solution.)

LucioFranco:
So I think the real goal here should be to make `Parts` cloneable and then since bodies are generic let bodies implement clone if they can. For streams, it should be possible to detect if the request failed before the stream was extracted to retry it then. The big issue for me is making Request a bit easier to deal with and extensions seem to always be the big blocker here.

davidbarsky:
I've shared my opinion on `extensions` previously and it still hasn't changed. Short version: I think should be removed. Long version: 

- There are three logical parts of an `http::Request`:
[...]

PR #574, opened 2022-11-15
Add `Clone` for Request/Response/Extensions

LucioFranco:
# Proposal

Allow `Request<T>`/`Respnose<T>` to implement `Clone if T: Clone`. The current blocker for this is that `Extensions` doesn't implement `Clone`. This is due to the complicated nature of `Any` and `Clone`. Though it is a solvable problem shown by `dyn-clone` and the `anymap` (beta releases) crates.

The solution in this PR swaps out our own `AnyMap` implementation to use the `anymap` crate `1.0.0-beta.2` version that supports clonable values. The downside to this change is that it nows requires all the types that get inserted into the anymap to implement clone which is not a backwards compatible change and will break end users. To support this use case this PR adds a `NotCloneExtension` (pending name, open to ideas) that allows users to implement Clone for a type T that doesn't implement Clone. This is done by storing the inner type as an Option and when cloning the newtype it will set the value to None. This allows users to for example, have different extension instances per requests in the retry middleware.  

The benefit of `Request<T>`/`Response<T>` implementing clone will greatly improve the ergonomics for end users and simplify things like the tower retry layer.

Closes https://github.com/hyperium/http/issues/395

cc @seanmonstar @hawkw @davidbarsky @hlbarber @LukeMathWalker 

# Notes

This PR uses a forked version of the anymap crate that adds a specialized `extend_map` fn. The plan I think going forward would be to vendor the minimal set of code from the anymap crate and to not depend on it for 1.0 of http.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,8 @@ bytes = "1"
 fnv = "1.0.5"
 itoa = "1"
 
+anymap = "1.0.0-beta.2"
+
 [dev-dependencies]
 indexmap = "<=1.8"
 quickcheck = "0.9.0"
@@ -62,3 +64,6 @@ path = "benches/method.rs"
 [[bench]]
 name = "uri"
 path = "benches/uri.rs"
+
+[patch.crates-io]
+anymap = { git = "https://github.com/LucioFranco/anymap" }
--- a/src/extensions.rs
+++ b/src/extensions.rs
@@ -1,9 +1,8 @@
-use std::any::{Any, TypeId};
-use std::collections::HashMap;
+use anymap::{CloneAny, Map};
 use std::fmt;
-use std::hash::{BuildHasherDefault, Hasher};
+use std::hash::Hasher;
 
-type AnyMap = HashMap<TypeId, Box<dyn Any + Send + Sync>, BuildHasherDefault<IdHasher>>;
+type AnyMap = Map<dyn CloneAny + Send + Sync>;
 
 // With TypeIds as keys, there's no need to hash them. They are already hashes
 // themselves, coming from the compiler. The IdHasher just holds the u64 of
@@ -31,7 +30,7 @@ impl Hasher for IdHasher {
 ///
 /// `Extensions` can be used by `Request` and `Response` to store
 /// extra data derived from the underlying protocol.
-#[derive(Default)]
+#[derive(Default, Clone)]
 pub struct Extensions {
     // If extensions are never used, no need to carry around an empty HashMap.
     // That's 3 words. Instead, this is only 1 word.
@@ -59,16 +58,10 @@ impl Extensions {
     /// assert!(ext.insert(4u8).is_none());
     /// assert_eq!(ext.insert(9i32), Some(5i32));
     /// ```
-    pub fn insert<T: Send + Sync + 'static>(&mut self, val: T) -> Option<T> {
+    pub fn insert<T: Clone + Send + Sync + 'static>(&mut self, val: T) -> Option<T> {
         self.map
-            .get_or_insert_with(|| Box::new(HashMap::default()))
-            .insert(TypeId::of::<T>(), Box::new(val))
-            .and_then(|boxed| {
-                (boxed as Box<dyn Any + 'static>)
-                    .downcast()
-                    .ok()
-                    .map(|boxed| *boxed)
-            })
+            .get_or_insert_with(|| Box::new(AnyMap::new()))
+            .insert(val)
     }
 
     /// Get a reference to a type previously inserted on this `Extensions`.
@@ -83,11 +76,8 @@ impl Extensions {
     ///
     /// assert_eq!(ext.get::<i32>(), Some(&5i32));
     /// ```
-    pub fn get<T: Send + Sync + 'static>(&self) -> Option<&T> {
-        self.map
-            .as_ref()
-            .and_then(|map| map.get(&TypeId::of::<T>()))
-            .and_then(|boxed| (&**boxed as &(dyn Any + 'static)).downcast_ref())
+    pub fn get<T: Clone + Send + Sync + 'static>(&self) -> Option<&T> {
+        self.map.as_ref().and_then(|map| map.get::<T>())
     }
 
     /// Get a mutable reference to a type previously inserted on this `Extensions`.
@@ -102,11 +92,8 @@ impl Extensions {
     ///
     /// assert_eq!(ext.get::<String>().unwrap(), "Hello World");
     /// ```
-    pub fn get_mut<T: Send + Sync + 'static>(&mut self) -> Option<&mut T> {
-        self.map
-            .as_mut()
-            .and_then(|map| map.get_mut(&TypeId::of::<T>()))
-            .and_then(|boxed| (&mut **boxed as &mut (dyn Any + 'static)).downcast_mut())
+    pub fn get_mut<T: Clone + Send + Sync + 'static>(&mut self) -> Option<&mut T> {
+        self.map.as_mut().and_then(|map| map.get_mut::<T>())
     }
 
     /// Remove a type from this `Extensions`.
@@ -122,16 +109,8 @@ impl Extensions {
     /// assert_eq!(ext.remove::<i32>(), Some(5i32));
     /// assert!(ext.get::<i32>().is_none());
     /// ```
-    pub fn remove<T: Send + Sync + 'static>(&mut self) -> Option<T> {
-        self.map
-            .as_mut()
-            .and_then(|map| map.remove(&TypeId::of::<T>()))
-            .and_then(|boxed| {
-                (boxed as Box<dyn Any + 'static>)
-                    .downcast()
-                    .ok()
-                    .map(|boxed| *boxed)
-            })
+    pub fn remove<T: Clone + Send + Sync + 'static>(&mut self) -> Option<T> {
+        self.map.as_mut().and_then(|map| map.remove::<T>())
     }
 
     /// Clear the `Extensions` of all inserted extensions.
@@ -166,9 +145,7 @@ impl Extensions {
     /// ```
     #[inline]
     pub fn is_empty(&self) -> bool {
-        self.map
-            .as_ref()
-            .map_or(true, |map| map.is_empty())
+        self.map.as_ref().map_or(true, |map| map.is_empty())
     }
 
     /// Get the numer of extensions available.
@@ -184,28 +161,26 @@ impl Extensions {
     /// ```
     #[inline]
     pub fn len(&self) -> usize {
-        self.map
-            .as_ref()
-            .map_or(0, |map| map.len())
+        self.map.as_ref().map_or(0, |map| map.len())
     }
 
     /// Extends `self` with another `Extensions`.
     ///
     /// If an instance of a specific type exists in both, the one in `self` is overwritten with the
     /// one from `other`.
-    /// 
+    ///
     /// # Example
-    /// 
+    ///
     /// ```
     /// # use http::Extensions;
     /// let mut ext_a = Extensions::new();
     /// ext_a.insert(8u8);
     /// ext_a.insert(16u16);
-    /// 
+    ///
     /// let mut ext_b = Extensions::new();
     /// ext_b.insert(4u8);
     /// ext_b.insert("hello");
-    /// 
+    ///
     /// ext_a.extend(ext_b);
     /// assert_eq!(ext_a.len(), 3);
     /// assert_eq!(ext_a.get::<u8>(), Some(&4u8));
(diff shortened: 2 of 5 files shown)


src/extensions.rs
```diff
@@ -31,7 +30,7 @@ impl Hasher for IdHasher {
 ///
 /// `Extensions` can be used by `Request` and `Response` to store
 /// extra data derived from the underlying protocol.
-#[derive(Default)]
+#[derive(Default, Clone)]
```
hlbarber, line 33: The `derive(Clone)` here means that the `NotCloneExtension` will remain in the new `AnyMap`, as `NotCloneExtension(None)`. Is this what we want?

An alternative here might be:

- The `HashMap` store values as `enum Value { Clone(Box<dyn CloneAny>), NotClone(Box<dyn Any>) }`,
- `Clone` filters out the `Value::NotClone`,
- Have two methods `fn insert` and `fn insert_clone`, only the latter requiring `Clone`.

hlbarber (commented):
Alternatives:

- Wrap the `dyn Any` in an `Arc` instead of `Box`, forgo the `Extensions::get_mut` method. We could provide a `Extensions::swap` instead. Perhaps it's possible to emulate `get_mut` by doing something akin to `Arc::make_mut`.
- The `Extensions::clone()` returns `Extensions::new()`.

Neither of them are compelling.

LucioFranco:
> Wrap the dyn Any in an Arc instead of Box, forgo the Extensions::get_mut method. We could provide a Extensions::swap instead. Perhaps it's possible to emulate get_mut by doing something akin to Arc::make_mut.

I think it makes more sense to let users add the Arc themselves rather than force it upon them? What if they want to use a Rc?

> The Extensions::clone() returns Extensions::new().

This just feels misleading and doesn't support an extension existing through the clones.

src/extensions.rs
LucioFranco: > The derive(Clone) here means that the NotCloneExtension will remain in the new AnyMap, as NotCloneExtension(None). Is this what we want?

Im not totally following what you mean by this?

hlbarber:
```rust
let mut extensions = Extensions::new();
extensions.insert(NotCloneExtension::new(3_u32));
let extensions_b = extensions.clone();
let value = extensions_b.get::<NotCloneExtension<u32>>();
```

`value` here is `Some(NotCloneExtension(None))` from what I read from the implementation. Is this what we want? If we take the changes described above and provide a hand written `Clone` implementation we could filter out the non-`Clone` extensions.

Having two separate methods `fn insert` and `fn insert_clone` then storing the `enum Value`, has the additional benefit of:

- It's an addition rather than a breaking change.
- It doesn't require a newtype when using `get` - e.g. `get::<T>` will retrieve `T` when it's inserted via `insert` or `insert_clone`. This is in contrast to `get::<NotCloneExtension<T>>` and `get::<T>`.

hlbarber: Something like this is what I was imagining

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=318f42c9a2477340c76ddf52aeefd908

It doesn't compile yet, but I think it would if we employed some more of the `anymap` tricks?

LukeMathWalker:
Is separating (or unbundling, if you prefer) `Extensions` from `Request`/`Response` an option on the design table? Or is that considered to be a no-starter?

LucioFranco:
@LukeMathWalker I don't think that changes much since we still need a single type to pass through. One option, would be to change tower to accept something like `request, context` but that doesn't really feel like an improvement over just making extensions cloneable. So I would say removing it doesn't make much sense right now and probably not worth exploring.

davidbarsky:
I think that making extensions cloneable is a good thing, but I think I agree with @LukeMathWalker. However, I don't think changing Tower to accept `(request, context)` is the right move, but a wrapper along the lines of `Context<Request>` _could be_, where `Context` can be provided by Tower and acts as a semi-smart pointer.

Before moving on, I wanted to add some notes on my interest in this feature:
- I wanted extensions to to be cloneable so that retries could be easier to implement. However, with the increasing adoption of http/2 and http/3, I'm not that is as important as it once was (at least _to me_), since I'm not sure request-level retries make sense when streaming bodies are a thing. Additionally, with a bit more reflection, I don't think making retries easy for http/1.1 is necessarily a good thing either: I think it makes overloading downstream services _way_ too easy and users should instead be pushed to use well-tuned retry middleware from libraries like Tower.
-  I don't really have data to back up this assertion—I'm going off entirely on vibes and aesthetics—but the inclusion of extensions inside of `http` felt a little odd to me. Extensions seems like a use-case better served by a crate aimed at more advanced users, such as Tower, but I can see how the convenience of extensions existing directly on `http::Request`/`http::Response` is useful.

In any case, I don't feel too strongly about the existence of extensions inside of `http` and I think that making extensions cloneable is _probably_ a net win, but if I had a vote, I think exploring a world where Tower a context-style wrapper to requests and response is absolutely warranted.

LucioFranco:
> since I'm not sure request-level retries make sense when streaming bodies are a thing.

Connection level/before streaming body failures are generally a thing you want to retry. 

>  I think it makes overloading downstream services way too easy and users should instead be pushed to use well-tuned retry middleware from libraries like Tower.

The plan is to implement good defaults in the batteries included retry middleware to prevent this stuff. I think no matter what we want to support retries and let downstream consumers decide if they want to enable it or not.

> Extensions seems like a use-case better served by a crate aimed at more advanced users, such as Tower, but I can see how the convenience of extensions existing directly on http::Request/http::Response is useful.

I think the ergonomics of it existing with one of the most used types in the ecosystem has a huge benefit. I think generally speaking we've been pretty happy with extensions and I have not seen many users complain about it. I think this extensions style has been successfully adopted by the aws sdk as well. 

I don't think doing something like `Context<Request>` is as ergonomic as just having a `Request`, especially in the space of something like tower where generics are already complicated. I think overall with our infrastructure we should aim to reduce complexity and embedding extensions imo is one way to get there. 

I think another thing we need to think about here is timelines and what makes sense. I don't know if anyone has the time to explore changing up context in tower and the deadline for a http 1.0 is fast approaching. I think we may want to consider keeping things how they are and making a change like this just improves some flexibility while not losing out on the benefits of the previous implementation.

jdisanti:
Our `SdkBody` that we use as the generic arg for `http::Request` is conditionally cloneable depending on its source. An `SdkBody` that wraps a user-provided stream isn't cloneable since the source is gone after its read, but a local file system file where we still have the path is cloneable. Given this, I don't know how much we would be able to take advantage of `http::Request` being `Clone` when `B: Clone`. We need some form of optionality that isn't in the type system, unless we greatly overhaul the SDK to use multiple body types, but that would make it harder to use I think.

> I think this extensions style has been successfully adopted by the aws sdk as well.

We use a similar property bag pattern in the SDK, but we ended up wrapping `http::Request` in our own `operation::Request` so that we could have cloneable properties as well as add a `try_clone` method that does some [heavy lifting of manually cloning `http::Request`](https://github.com/awslabs/smithy-rs/blob/fba88df3eeb6c5ea8be2ea2df57232366c016b37/rust-runtime/aws-smithy-http/src/operation.rs#L199).

hawkw:
Regarding moving the extensions out to a separate type...it's worth noting that much of the value from having a notion of "extensions" comes from it being defined in the `http` crate, so that different crates that might pass around an `http::Request`/`http::Response` can propagate the data that other libraries want to associate with that request/response, without having to be _aware_ of those other libraries.

For example, `hyper`'s implementation of protocol upgrades (e.g. `CONNECT`) relies on setting request/response extensions: https://github.com/hyperium/hyper/blob/96dcc79b62572c591b8062f26db602868de5b5f0/src/upgrade.rs#L304 

Because the extensions are part of the `http::Request` type, `hyper` can rely on the data it stores for upgrades being propagated by other libraries involved in handling a request, such as `tower` or web frameworks built on top of `hyper`. 
If the solution was that crates like `tower-http` or web frameworks using `hyper` should define their own types that store extensions, then `hyper` would have to be aware of those types...which is untenable for several reasons (e.g. `hyper` needing only stable dependencies, cyclic deps between web frameworks that depend on `hyper` and define their own context types, `hyper` potentially having to implement separate code paths for each dependent's context type, etc).

AFAICT, a model where other libraries provide the context/extension type wrapping a `Request` kind of defeats the purpose of having a typemap of extensions in the first place, since they can no longer rely on their extensions being propagated across library boundaries when `http` is the only dependency shared by those libraries. The entire reason we store extensions in a typemap is to erase types so that they can cross crate boundaries without requiring every crate to be aware of a specific extension type. If we decide to give up on having that functionality in `http`, crates that want to wrap requests could just have typed structs that store their extensions as fields, and we wouldn't need the typemap...but we would be losing the ability to propagate extensions through libraries that are unaware of any given extension.

davidbarsky:
(i'm little sick, so I apologize if this response isn't particularly coherent or well-edited.)

> Connection level/before streaming body failures are generally a thing you want to retry.

Fair enough!

> The plan is to implement good defaults in the batteries included retry middleware to prevent this stuff. I think no matter what we want to support retries and let downstream consumers decide if they want to enable it or not.

I had a longer response, but after re-reading this, I'm probably over-indexing on this point and was probably steeped too long in the context of Thrift-based RPC, so I replaced it with this sentence.

> I think the ergonomics of it existing with one of the most used types in the ecosystem has a huge benefit. I think generally speaking we've been pretty happy with extensions and I have not seen many users complain about it. I think this extensions style has been successfully adopted by the aws sdk as well.

Sure, I can see that.

> I don't think doing something like Context<Request> is as ergonomic as just having a Request, especially in the space of something like tower where generics are already complicated. I think overall with our infrastructure we should aim to reduce complexity and embedding extensions imo is one way to get there.

🤷. I think I've had stronger opinions in the past but they're not so strong now. I think most of my complaints boil down to aesthetics and vibes, which isn't really the best basis for making decisions like these.

> I think another thing we need to think about here is timelines and what makes sense. I don't know if anyone has the time to explore changing up context in tower and the deadline for a http 1.0 is fast approaching. I think we may want to consider keeping things how they are and making a change like this just improves some flexibility while not losing out on the benefits of the previous implementation.

I haven't really been keeping pace with Hyper's roadmap to 1.0, so I'm assuming it's trying to hit 1.0 this year, which: fair.

> For example, hyper's implementation of protocol upgrades (e.g. CONNECT) relies on setting request/response extensions: https://github.com/hyperium/hyper/blob/96dcc79b62572c591b8062f26db602868de5b5f0/src/upgrade.rs#L304

I keep forgetting about Hyper's usage of extensions in upgrades; I think that usecase is sufficiently compelling that it overrides any aesthetic concerns I might have.

>  The entire reason we store extensions in a typemap is to erase types so that they can cross crate boundaries without requiring every crate to be aware of a specific extension type. If we decide to give up on having that functionality in http, crates that want to wrap requests could just have typed structs that store their extensions as fields, and we wouldn't need the typemap...but we would be losing the ability to propagate extensions through libraries that are unaware of any given extension.

That is a compelling motivation and I think the reason I forgot about this is that I haven't been steeped in writing http-based services (or _any_ services, for that matter) for nearly the last two years.

In any case, I trust y'all will make the right decision and this PR is an improvement to the status quo.

LukeMathWalker:
> Regarding moving the extensions out to a separate type...it's worth noting that much of the value from having a notion of "extensions" comes from it being defined in the http crate, so that different crates that might pass around an http::Request/http::Response can propagate the data that other libraries want to associate with that request/response, without having to be aware of those other libraries.

To clarify: I was not suggesting to remove `http::Extension` - having a "standard" extension map type that everybody uses is extremely advantageous for all the reasons you enumerated.
My suggestion was much more limited: keep `http::Extension`, but do not store it as a field in `http::Request`/`http::Response`.

It would open up a few possibilities for crates that depend on `http` (e.g. I've often seen users confused when they found out that something they inserted in the extension map of an `http::Request` doesn't show up in the extensions map for the returned `http::Response`) but it does come with an impact on ergonomics - you know have to pass two things around instead of one, which can be annoying.

benwis:
Was wondering if there was any movement on this? Making Request Clone would solve some of our problems for use of the http Request and Response types

seanmonstar (changes requested):
@LucioFranco consensus in the issue was to move forward with this. Would you be able to update the PR?

Cargo.toml
```diff
@@ -28,6 +28,8 @@ bytes = "1"
 fnv = "1.0.5"
 itoa = "1"
 
+anymap = "1.0.0-beta.2"
```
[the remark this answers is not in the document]
LucioFranco: Yes, that was my initial goal but to poc I just used it and as you can see its patched so we probably want to vend anyways.

benwis:
I am irrationally excited this might become get merged. @LucioFranco let me know if I can help

seanmonstar:
This was done in #395

PR #576, opened 2022-11-15
chore: Update msrv to 1.56

LucioFranco:


LucioFranco:
@seanmonstar I think we just need to remove the requirement for the 1.49 test and change it to 1.56

seanmonstar:
> I think we just need to remove the requirement for the 1.49 test and change it to 1.56

What do you mean? Do you mean change the MSRV CI job to only do `cargo build` instead of `cargo test`?

LucioFranco:
No I mean on github actions if you look below its waiting for a required 1.49 test which is no longer 1.49 but 1.56. This needs to be changed in branch protection settings.

seanmonstar:
Oh. Would what I mentioned be true, though? The crate itself still builds on 1.49, it's just some test dependencies?

hawkw:
If 1.56.0 is only required for benches, it definitely seems preferable to continue supporting 1.49.0 unless we actually want to use features from 1.56.0 in the crate itself?

PR #578, opened 2022-11-27
Add HeaderMap collect from array

light4:
Act like [std::collections::HashMap][std_collections_from_array] from std from 1.56.0

[std_collections_from_array]: https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#impl-From%3C%5B(K%2C%20V)%3B%20N%5D%3E-for-HashMap%3CK%2C%20V%2C%20RandomState%3E

### DIFF
--- a/src/header/map.rs
+++ b/src/header/map.rs
@@ -1813,6 +1813,12 @@ impl<T> IntoIterator for HeaderMap<T> {
     }
 }
 
+impl<T, const N: usize> From<[(HeaderName, T); N]> for HeaderMap<T> {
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for the PR! However, I don't think having this sugar in the library is what we want. There is the existing FromIterator implementation that helps with this.

light4:
This PR makes it more consistent with rust std.
Maybe take another look at https://github.com/rust-lang/rust/pull/84111

PR #579, opened 2022-12-01
chore: Fix MSRV CI test

LucioFranco:


.gitignore
```diff
@@ -1,2 +1,3 @@
 target
 Cargo.lock
+.DS_Store
```
seanmonstar, line 3: Can we not include these files?

LucioFranco: yeah let me back this out

seanmonstar: And the file in the `benches` directory (I can't comment on it directly).

seanmonstar (approved):

PR #581, opened 2022-12-07
feat: Add one Grammatical sugar

tu6ge:
Examples:

```
use http::HeaderMap;

fn hello<M: Into<HeaderMap>>(map: M){
    let map: HeaderMap = map.into();
    // more ...
}

// before call it with zero item
use http::HeaderMap;
hello(HeaderMap::new());
// now call it with zero item
hello([]);

```


tu6ge:
It suddenly dawned on me, `FromIterator` is a more abstract `From` type. Wonderful rust !

Issue #573, opened 2022-11-08
Define policy for which header name constants to include


seanmonstar:
We include many standard header names as constants in `http::header`. The exact criteria has not been defined, and it would be wise to do so. This stops it from being an unwritten rule in a few people's heads. The policy likely should consider the pros of adding any name versus the costs of adding a name.

seanmonstar:
Thanks for listing those, it may be worth considering quicker if we should chop out some of the obscure ones that are currently in the crate before 1.0 (which is coming soon!).

acfoltzer:
A good place to start is the IANA Field Name Registry. The registry itself is [here](https://www.iana.org/assignments/http-fields/http-fields.xhtml#field-names), and the specification of the registry is [here](https://www.rfc-editor.org/rfc/rfc9110.html#name-field-name-registry) in RFC 9110.

In [this gist](https://gist.github.com/acfoltzer/9204fe6e9781f0a9a01105b44881f1e1) I compared the names in the registry vs the names defined as constants in `http`. There are quite a few more in the registry than we currently have as constants. I'm in favor of including any registry headers that aren't `provisional` as constants in this crate, as they are very unlikely to ever go away.

There are three sets that I think are more debatable.

## `provisional` headers in the IANA registry

```
amp-cache-transform
configuration-context
ediint-features
isolation
repeatability-client-id
repeatability-first-sent
repeatability-request-id
repeatability-result
sec-gpc
timing-allow-origin
```

I'm inclined to not proactively make constants for these given their lack of stability, but we should be open to their inclusion if folks request them and it appears they're on track to standardize.

## Header constants currently in `http` but not the registry

```
dnt
referrer-policy
upgrade-insecure-requests
x-dns-prefetch-control
x-xss-protection
```

These probably need to be handled on a case-by-case basis, as they might be referenced in other specs. For example, `dnt` is part of the [Fetch standard](https://fetch.spec.whatwg.org/#forbidden-request-header). 

## Headers in neither the registry nor `http`

...but present in widely-cited sources such as MDN. For example, [`Accept-CH-Lifetime`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-CH-Lifetime). These probably also need to be treated on a case-by-case basis.

seanmonstar:
[...]

PR #583, opened 2023-01-20
feat: add cache-status and cdn-cache-control header name constants

acfoltzer:
Docs include pointers to the relevant RFCs, as these are not in the base HTTP semantics spec.


seanmonstar:
I think this issue would be relevant here: https://github.com/hyperium/http/issues/573

acfoltzer:
@seanmonstar thanks for the tip. I made some recommendations over in that thread. Since both of these are part of the IANA registry as non-provisional entries, I think we should definitely include them, but perhaps we should should add all the non-provisional registry names in one go?

seanmonstar:
I was trying to read up on these headers, I didn't remember what they were. That's because they're new (2022)! Their respective RFCs claim their current status is "Proposed Standard". Are they used elsewhere? I don't seem to notice them on sites behind Cloudflare, Cloudfront, or GCP.

acfoltzer:
Yeah, they're still relatively new, but are a standardization of a scattering of implementation-defined headers like `Surrogate-Control` and `X-Cache`. FWIW, once something makes it to the `Proposed Standard` state, things like header names are very unlikely to change (thus the `permanent` status for these in the IANA field name registry), it's mostly about nailing down details of edge cases at that point.

seanmonstar (approved):

PR #584, opened 2023-01-24
double quote "Sensitive" header value

haruska:
In the implementation of `Debug` for `HeaderValue` if a value is not marked sensitive, the value is written to the formatter with double quotes added. However, if marked sensitive the unquoted static string `Sensitive` is instead written to the formatter. This causes issues when attempting to output headers during tracing/logging.

This PR adds double quotes to the "Sensitive" string output in the case of a sensitive `HeaderValue`.

### DIFF
--- a/src/header/value.rs
+++ b/src/header/value.rs
@@ -368,7 +368,7 @@ impl AsRef<[u8]> for HeaderValue {
 impl fmt::Debug for HeaderValue {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         if self.is_sensitive {
-            f.write_str("Sensitive")
+            f.write_str("\"Sensitive\"")
         } else {
             f.write_str("\"")?;
             let mut from = 0;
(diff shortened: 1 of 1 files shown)


seanmonstar:
It would be interesting to hear what issues that causes. The existing implementation purposefully doesn't wrap the word in quotes, so as to not be confused with a value that literally has the string "Sensitive".

haruska:
> The existing implementation purposefully doesn't wrap the word in quotes

That's fair. My specific issue was using [tower-http DefaultMakeSpan to output http headers](https://docs.rs/tower-http/latest/src/tower_http/trace/make_span.rs.html#65-68).  Specifically [Line 91](https://docs.rs/tower-http/latest/src/tower_http/trace/make_span.rs.html#91) adds a `headers` key with the value being the map of http headers. The header values get outputted with double quotes.

This works fine until a header value gets marked sensitive. Then, that header value is output as Sensitive (without double quotes) while other values are quoted. Unfortunately, I'm outputting as JSON which makes the Sensitive output invalid.

The answer could just be I need to write a serializer for the headers which check the same sensitive flag and set a known string in the JSON output. However, it might make more sense (or be less confusing for other devs) for the tower-http default MakeSpan to just continue to work with marked sensitive values.

Maybe some more obvious string like "<REDACTED>"? It would just be nice to have the output always be quoted strings.

PR #586, opened 2023-01-27
Add `Display` for Version

emanuele-em:
Add `Display` for Version


emanuele-em (commented):
- [ ]

seanmonstar (changes requested):
We purposefully don't implement it for Version, because a semantic display doesn't make much sense.

emanuele-em:
I implemented it because it would be convenient to have the to_string that exists for the status, the method etc...

PR #588, opened 2023-02-09
status: add CONTENT_TOO_LARGE and UNPROCESSABLE_CONTENT aliases

mina86:
RFC9110 changed phrases for status code 413 and 422:
* 413 Payload Too Large → Content Too Large
* 422 Unprocessable Entity → Unprocessable Content

Introduce CONTENT_TOO_LARGE and UNPROCESSABLE_CONTENT StatusCode const
items to reflect those changes and update phrases used in those status
code.

While at it, update phrase for status 203 to ‘Non-Authoritative
Information’ (with a dash) and update all links to go to RFC 9110.


src/status.rs
```diff
@@ -298,19 +298,32 @@ macro_rules! status_codes {
...
+            }
+
         $(
             $(#[$docs])*
-            pub const $konst: StatusCode = StatusCode(unsafe { NonZeroU16::new_unchecked($num) });
+            pub const $name: StatusCode = StatusCode::from_u16_or_panic($num);
+            $(
+                #[doc = concat!("Alias of [`", stringify!($name),
+                                "`](`Self::", stringify!($name), "`).")]
+                pub const $alias: StatusCode = StatusCode::$name;
```
seanmonstar, line 318: Should these be deprecated now, and removed for http 1.0?

mina86: I dunno.  Status code phrases does really matter and deprecating those names just seems like an unnecessary burden for users.  It’s not like keeping those names is considerable maintenance burden.

seanmonstar: Fair point, thanks for entertaining my random musings :)

mina86:
Friendly ping.

Issue #377, opened 2020-01-02
allow camel case of header name


xudesheng:
I'm working on a customized websocket protocol which is using a header name "appKey" for authentication purpose. However, all customized header name is lower cased in this crate. Is there a public API which can enable acceptance of camel case of header name and don't auto lower case it?

I'm trying to use actix-ws, which is relying on this crate.

carllerche:
My understanding is that case sensitive HTTP header names is invalid usage. This might need to be handled outside of `http`.

Darkspirit:
According to HTTP standards your app must compare header names case-insensitively. Otherwise you will definitely run into compatibility problems.

seanmonstar:
For more context, see this previous issue: https://github.com/hyperium/http/issues/228

Issue #228, opened 2018-07-14
X-<Header> converts to lowercase


crusty-dave:
With the following code:

`Ok(HttpResponse::Ok() .header("X-Subject-Token", token.token) .body(""))`

It transmits "X-Subject-Token" as "x-subject-token", this would seem to be a bug, it shouldn't be changing any header tags like this AFAIK?

Per RFC 7230, the headers should be case-insensitive, but I would still expect http to honor the case that I have provided.

[RFC 7230](https://tools.ietf.org/html/rfc7230#page-19)

> 3.2. Header Fields
> 
> Each header field consists of a case-insensitive field name followed
> by a colon (":"), optional leading whitespace, the field value, and
> optional trailing whitespace.
> 

Note that the HTTP RFCs use camel-case with dash (-) separators when they describe the fields.

This is the first time that I have witnessed something like "content-type" instead of "Content-Type".

I would think that this might cause some compatibility issues with some http processors if they ignored section 3.2 (I ran into some code with that issue).  So at the least, there should be an option to preserve Camel-Case if present IMO.

carllerche:
Thanks for the report.

The HTTP crate internally normalizes to lower case in order to implement a number of performance optimizations. As you quoted, the spec defines HTTP header fields as being case-insensitive.

The `http` crate has no plans to add support for maintaining header name case as this would introduce a performance regression for all users.

It is possible for HTTP client and server implementations to add features supporting invalid HTTP peers without the `http` crate doing anything. If you hit a real interopt problem, I would suggest opening an issue with the client / server you are using.

I'm going to close this issue as there is no further action for us to take.

seanmonstar:
As you've pointed out, RFC 7230 says names are case-insensitive, so *should* be perfectly valid to treat `X-Subject-Token` as `x-subject-token`. Additionally, in HTTP/2, **all** header names are lowercase (it's specified to be a protocol error for upper case names). Because of this, `HeaderName` chooses to always handle names as lower case. Doing this allows optimizations: case-insensitive hashing and comparison are no longer required, improving speed of inserts and searches in the `HeaderMap`.

[...]

PR #593, opened 2023-03-06
Allow mixed case header names.

abamaxa:
I've run into a problem with the browser installed on recent Samsung televisions where HTTP headers that are entirely lower case are not recognized, preventing the video control from working correctly. 

This PR adds a new method `HeaderName::from_static_preserve_case` that creates a header name with the same case as the string passed to it, e.g.

`HeaderName::from_static_preserve_case("Content-Range")` will produce a header named `Content-Range` instead of `content-range`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,9 +18,6 @@ jobs:
           - stable
           - beta
           - nightly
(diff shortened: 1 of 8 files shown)


seanmonstar:
The header names are purposefully lowercase in this library. It's been discussed here: #228 #377

Issue #587, opened 2023-02-01
A `method!` macro  to define methods at build time


WhyNotHugo:
Using [HTTP verbs / methods that are not defined in this crate](https://github.com/hyperium/http/issues/346) has very ugly semantics, e.g.:

```rust
Method::from_bytes(b"PROPFIND").expect("PROPFIND is a valid method")
```

This `expect` adds a lot of noise all over the place, but it also can't be defined as a `const`. The following in invalid:

```rust
const PROPFIND: Method = Method::from_bytes(b"PROPFIND")
  .expect("PROPFIND is a valid method")
```

At the same time, these verbs are something that can be validated at build time (rather than runtime). Do you think a macro to create methods would be reasonable? I'm thinking something along the lines of:

```rust
method!("PROPFIND");
```

It would simply be replaced by the correct method, but a `proc_macro` could allow validating that the string is an acceptable verb at build time (hence, avoiding having a `Result` for an operation that is known to be infallible).

seanmonstar:
I imagine a `from_static` could just be made a `const fn`, like done for headers.

WhyNotHugo:
I tried making it `const fn`. I also had to make functions that it calls `const fn` and replace some `?` with `match {...}`.

Finally, I reached `AllocatedExtension::new`, where I'm not sure how to continue:

```diff
diff --git a/src/method.rs b/src/method.rs
index b7b3b35..a795208 100644
--- a/src/method.rs
+++ b/src/method.rs
@@ -67,7 +67,6 @@ enum Inner {
     ExtensionAllocated(AllocatedExtension),
 }
 
-
 impl Method {
     /// GET
     pub const GET: Method = Method(Get);
@@ -97,7 +96,7 @@ impl Method {
     pub const TRACE: Method = Method(Trace);
 
     /// Converts a slice of bytes to an HTTP method.
-    pub fn from_bytes(src: &[u8]) -> Result<Method, InvalidMethod> {
+    pub const fn from_bytes(src: &[u8]) -> Result<Method, InvalidMethod> {
         match src.len() {
             0 => Err(InvalidMethod::new()),
             3 => match src {
@@ -128,7 +127,10 @@ impl Method {
                 if src.len() < InlineExtension::MAX {
                     Method::extension_inline(src)
                 } else {
-                    let allocated = AllocatedExtension::new(src)?;
+                    let allocated = match AllocatedExtension::new(src) {
+                        Ok(a) => a,
+                        Err(e) => return Err(e),
[...]

PR #595, opened 2023-03-27
Add Method::from_static

WhyNotHugo:
Allows creating constant `Method` instances, e.g.:

    const PROPFIND: Method = Method::from_static(b"PROPFIND");

Fixes: https://github.com/hyperium/http/issues/587


src/method.rs
```diff
@@ -440,6 +475,17 @@ mod test {
         assert_eq!(Method::GET, &Method::GET);
     }
 
+    #[test]
+    fn test_from_static() {
```
seanmonstar, line 479: Should probably also test that `assert_eq!(Method::from_static(b"GET"), Method::GET)`.

src/method.rs
```diff
@@ -335,6 +347,29 @@ mod extension {
...
+        /// Convert static bytes into an `InlineExtension`.
+        ///
+        /// Only the first 15 characters are used.
+        ///
+        /// # Panics
+        ///
+        /// If the input bytes are not a valid method name.
+        pub const fn from_static(src: &'static [u8]) -> InlineExtension {
+            let mut i = 0;
+            let mut dst = [0u8;15];
```
seanmonstar, line 359: Looks like the loop doesn't do anything if the `src` is longer than 15, so would it just truncate? Probably worth having the unit tests include a much longer string passed to `from_static`.

src/method.rs
```diff
@@ -136,6 +136,18 @@ impl Method {
...
 
+    /// Convert static bytes into a `Method`.
+    ///
+    /// The method name can only be up to 15 bytes long.
+    ///
+    /// # Panics
+    ///
+    /// If the input bytes are not a valid method name.
+    pub const fn from_static(src: &'static [u8]) -> Method {
+        let inline = InlineExtension::from_static(src);
```
seanmonstar, line 147: Shouldn't it also check if it could one of the "standard" variants?

src/method.rs
```diff
@@ -440,6 +488,19 @@ mod test {
         assert_eq!(Method::GET, &Method::GET);
     }
 
+    #[test]
+    fn test_from_static() {
+        assert_eq!(Method::from_static(b"PROPFIND"), Method::from_bytes(b"PROPFIND").unwrap());
+        assert_eq!(Method::from_static(b"GET"), Method::from_bytes(b"GET").unwrap());
```
seanmonstar, line 494:
```suggestion
        assert_eq!(Method::from_static(b"GET"), Method::GET);
```

WhyNotHugo: I kept both examples just to be sure.

WhyNotHugo:
Oh, I just realised that CI failed due to `panic!` not being available on the MSRV. I'll try refactoring that.

seanmonstar:
We could probably update the MSRV, depending on what the version required is and how old it is.

WhyNotHugo:
For tokio, the current MSRV is 1.56.0.

WhyNotHugo:
I've updated the PR to panic without `panic!`. I used `([] as [u8; 0])[0];`, which is already used as a `const panic!` hack in a couple of other places in the codebase.

We can remove this hack (both this new instance and existing ones) once the MSRV is bumped to 1.57.0 or greater.

lperlaki:
I was wondering if we could add a variant `StaticRef(&'static [u8])` to the `Method` `Inner` enum instead of panicking on longer than 15 byte methods?

WhyNotHugo:
I don't know of any methods that are longer than 15 bytes, so I'm not sure that it's worth the effort.

2xsaiko:
What's this stuck on?

lperlaki:
I would also like to see this being merged as it helps with implementing a webdav server base on hyper


> I don't know of any methods that are longer than 15 bytes, so I'm not sure that it's worth the effort.

 recently I stubbled over the `BASELINE-CONTROL` webdav method wich is longe then 15 bytes (16)


> I was wondering if we could add a variant `StaticRef(&'static [u8])` to the `Method` `Inner` enum instead of panicking on longer than 15 byte methods?

figured out that this wouldn't work with the current implementation since the equality also checks the enum variant wich is required that `Method` can support pattern matching

orium:
@seanmonstar Anything that needs to be done to land this? Happy to help if there's anything to do.

WhyNotHugo:
> I was wondering if we could add a variant StaticRef(&'static [u8]) to the Method Inner enum instead of panicking on longer than 15 byte methods?

I explored this a bit further.

Adding a new enum variant won't increase the total size of `Inner`. The associated data is a single reference, so it's also not bigger than the largest of the existing variants.

I've implemented your suggestions. This remove the conditional (compile-time) panic if len > 15.

WhyNotHugo:
At this point, waiting for feedback from the core devs.

src/method.rs
```diff
@@ -134,6 +138,30 @@ impl Method {
...
+            b"POST" => Method::POST,
+            b"PUT" => Method::PUT,
+            b"DELETE" => Method::DELETE,
+            b"HEAD" => Method::HEAD,
+            b"TRACE" => Method::TRACE,
+            b"CONNECT" => Method::CONNECT,
+            b"PATCH" => Method::PATCH,
+            src => {
+                if src.len() <= 15 {
+                    let inline = InlineExtension::from_static(src);
```
seanmonstar, line 155: With an `ExtensionStatic` available, is there a benefit to keeping this one?

lperlaki:
I found the Problem with `ExtensionStatic` is that

```rust
const BASELINE_CONTROL: Method = Method::from_static(„BASELINE-CONTROL“);

let m = Method::from_bytes(„BASELINE-CONTROL“);

let is_baseline = match m {
    BASELINE_CONTROL => true,
    _ => false,
};
```

will not match.

src/method.rs
WhyNotHugo: `InlineExtension` has no heap allocation and has the data available in the same allocation. Most HTTP methods fit in this case, and it has no overhead.

`ExtensionStatic` is a niche for unusually long methods, but has a heap allocation and an extra indirection.

We have three options here:

- Keep both. No real downsides aside from another internal type, and does not increase the size of the `Method` type.
- Keep only `InlineExtension`, restricting `from_static` to only methods of `len <= 15`.
- Keep only `ExtensionStatic`, forcing any `from_static` method to use a heap allocation and an extra indirection.

Personally, I prefer the first choice the most, and the third one the least (since it introduces pointless overhead for most WebDAV methods, albeit minor).

WhyNotHugo:
> I found the Problem with `ExtensionStatic` is that
> 
> ```rust
> const BASELINE_CONTROL: Method = Method::from_static(„BASELINE-CONTROL“);
> 
> let m = Method::from_bytes(„BASELINE-CONTROL“);
> 
> let is_baseline = match m {
>     BASELINE_CONTROL => true,
>     _ => false,
> };
> ```
> 
> will not match.

I'll add a test for this case and figure out what's wrong.

lperlaki:
> * Re-implement `PartialEq` por `Inner`. This is likely a lot of very verbose code, and the derived implementation has optimisations which I suspect the core devs will want to keep around.

Not deriving `PartialEq` will also not implement `StructuralPartialEq` which is needed if we want to keep the `match` functionality

> * Drop `ExtensionInline(InlineExtension)`, which means that common short methods will always be behind a pointer dereference. Not a terrible performance drop, but given how much all the other variants are optimised, it seems non-ideal.

will not fix the `PartialEq` implementation between Static and Allocated variant if not by hand implemented and then we have the first problem again 

> * Drop `ExtensionStatic(StaticExtension)` entirely, and limit `from_static` to a maximum of 15 bytes, as originally implemented. This is probably the least ideal approach.

I think this is the only viable solution if we want to keep the pattern matching behaviour

PR #606, opened 2023-05-05
Implement the Early-Data (RFC 8470) header

### LINKED ISSUE #529: Early-Data header (RFC 8470)
[RFC 8470](https://datatracker.ietf.org/doc/html/rfc8470) defines an `Early-Data` header, but the `http::header` module doesn't have a corresponding `EARLY_DATA` constant that makes it convenient to use. Could such a thing be added, please?

dasJ:
Closes #529


dasJ:
Ping maintainers?

PR #613, opened 2023-06-12
Fix insufficient validation of the Uri authority component can lead to a panic when calling `host()` afterwards.

f0rki:
Previously only the presence of both square brackets was checked, but not the order. This made `]o[` validate as a good uri authority. However, calling the `Authority::host` function then panics.

I also added a check for too many `:` chars in the authority.


82marbag:
CI is failing, but the issue seems unrelated. @seanmonstar can you verify please?

seanmonstar:
Oh, yea, seems like a dev-dependency is past our MSRV. I've filed #614 to only run `cargo check`.

seanmonstar:
Ok, fixed that, if you want to rebase, it should be better here.

seanmonstar (approved):

PR #623, opened 2023-08-28
Implement From<PathAndQuery> for Bytes

WhyNotHugo:
This allows getting a `Bytes` instance from a `PathAndQuery` without any copying.


tesaguri:
This has already been tracked as part of #459.

WhyNotHugo:
Indeed, #470 seems to include this along with many other changes.

I'll leave this open for now; at least until there's some feedback for #470.

PR #624, opened 2023-09-13
Add sockaddr authority

mcr:
This adds the ability to create an authority object from a SocketAddr.

While many web interactions start with a URL, there are others that start from a mDNS, DNS-SD or GRASP (RFC8990) discovery.  IP addresses (often link-local), port numbers (and IPv6 scope) are returned from the discovery, but one then needs to form a URI for the relevant HTTP library.
(Although the application usually has to open the socket itself, a URI is often still required for Host: header, and path+query string)

Perhaps this should be hidden behind some feature to avoid a dependency upon net.
Probably this should also be behind no_std, since byte strings are created.

I was surprised that the URI was not parsed into pieces and pointed to separately, but I see the logic for this.
I would prefer to put this into authority because that keeps the knowledge about what are valid and invalid URIs in one place.
(Thank you, btw, for supporting RFC6874, and note draft-ietf-6man-rfc6874bis/ about to be published)

### DIFF
--- a/src/uri/authority.rs
+++ b/src/uri/authority.rs
@@ -547,6 +547,14 @@ mod tests {
     }
 
     #[test]
+    fn equal_to_self_of_same_authority_with_port() {
+        let authority1: Authority = "example.com:80".parse().unwrap();
+        let authority2: Authority = "EXAMPLE.COM:80".parse().unwrap();
+        assert_eq!(authority1, authority2);
+        assert_eq!(authority2, authority1);
+    }
+
+    #[test]
     fn not_equal_to_self_of_different_authority() {
         let authority1: Authority = "example.com".parse().unwrap();
         let authority2: Authority = "test.com".parse().unwrap();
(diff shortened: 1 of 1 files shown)


mcr:
@becarpenter might have some test cases to add.

becarpenter:
I don't have context to know what sort of test cases you would want. There are examples in [draft-ietf-6man-rfc6874bis](https://datatracker.ietf.org/doc/draft-ietf-6man-rfc6874bis/).
Re RFC6874(bis), note that the bis document removes the need to replace "%" by "%25". The impact of that depends on fine details of the URL parser. My fix for the WGET parser is at https://github.com/becarpenter/wget6, and the way WGET is coded means that it accepts both RFC6874 and RFC6874bis formats.

src/uri/authority.rs
```diff
@@ -68,6 +69,49 @@ impl Authority {
...
+    ///
+    /// This includes formatting the IPv4 or IPv6 into a string, adding the scope for IPv6 addresses,
+    /// enclosing it in [], and then adding any port number present.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use http::uri::Authority;
+    /// let authority = Authority::from_static("example.com");
+    /// assert_eq!(authority.host(), "example.com");
```
peddermaster2, line 85: This example is not for the function.

Issue #627, opened 2023-09-19
Arithmetic overflow found on `with_capacity()`


HeeillWang:
I executed fuzz testing on `http-0.2.9`, and found some arithmetic overflow.
Please note that overflow condition is different with #626 .

~~~
    pub fn with_capacity(capacity: usize) -> HeaderMap<T> {
        if capacity == 0 {
            HeaderMap {
                mask: 0,
                indices: Box::new([]), // as a ZST, this doesn't actually allocate anything
                entries: Vec::new(),
                extra_values: Vec::new(),
                danger: Danger::Green,
            }
        } else {
            let raw_cap = to_raw_capacity(capacity).next_power_of_two();   // overflow!
            assert!(raw_cap <= MAX_SIZE, "requested capacity too large");
            debug_assert!(raw_cap > 0);

            HeaderMap {
                mask: (raw_cap - 1) as Size,
                indices: vec![Pos::none(); raw_cap].into_boxed_slice(),
                entries: Vec::with_capacity(raw_cap),
                extra_values: Vec::new(),
                danger: Danger::Green,
            }
        }
    }
~~~

reproduce with : 
~~~
HeaderMap::<u32>::with_capacity(12538021362599493900);  // put some big number here
~~~
If you input TOO big number on `with_capacity()`, #626 occurs before reaching to `next_power_of_two()`.

hawkw:
It seems like there should be an assertion prior to the `next_power_of_two` call, in addition to the one after it?

HeeillWang:
That would resolve #626 as well.
Plus, consider to use [checked_next_power_of_two](https://doc.rust-lang.org/std/primitive.usize.html#method.checked_next_power_of_two)

seanmonstar:
Thanks for detecting those! Would you like to submit a PR that uses an assertion or checked math?

HeeillWang:
submitted on #628

PR #628, opened 2023-09-19
Resolve arith overflow on with_capacity

### LINKED ISSUE #626: Arithmetic overflow found on `to_raw_capacity()`
I executed fuzz testing on `http-0.2.9`, and found some arithmetic overflow.

~~~
#[inline]
fn to_raw_capacity(n: usize) -> usize {
    n + n / 3   // overflow!
}
~~~

reproduce with : 
~~~
HeaderMap::<u32>::with_capacity(17798225791641714878);  // put some big number here
~~~

HeeillWang:
resolve #626  #627

### DIFF
--- a/src/header/map.rs
+++ b/src/header/map.rs
@@ -472,7 +472,9 @@ impl<T> HeaderMap<T> {
                 danger: Danger::Green,
             }
         } else {
-            let raw_cap = to_raw_capacity(capacity).next_power_of_two();
+            let raw_cap = to_raw_capacity(capacity)
+                .checked_next_power_of_two()
+                .expect("requested capacity overflows on checked_next_power_of_two");
             assert!(raw_cap <= MAX_SIZE, "requested capacity too large");
             debug_assert!(raw_cap > 0);
 
@@ -3218,7 +3220,8 @@ fn usable_capacity(cap: usize) -> usize {
 
 #[inline]
 fn to_raw_capacity(n: usize) -> usize {
-    n + n / 3
+    n.checked_add(n / 3)
+        .expect("overflow while converting to raw capacity")
 }
 
 #[inline]


src/header/map.rs
```diff
@@ -472,7 +472,9 @@ impl<T> HeaderMap<T> {
                 danger: Danger::Green,
             }
         } else {
-            let raw_cap = to_raw_capacity(capacity).next_power_of_two();
+            let raw_cap = to_raw_capacity(capacity)
+                .checked_next_power_of_two()
+                .expect("requested capacity overflows on checked_next_power_of_two");
```
hawkw, line 477: tiny nit, not a big deal: what do you think about changing this `expect` to a `match` and `panic!`, so that the panic message can include the requested capacity?
```suggestion
            let raw_cap =  match raw_capacity(capacity).checked_next_power_of_two() {
                Some(c) => c,
                None => panic!("requested capacity {capacity} too large: next power of two would overflow `usize`"),
            };
```

src/header/map.rs
```diff
@@ -3218,7 +3220,8 @@ fn usable_capacity(cap: usize) -> usize {
 
 #[inline]
 fn to_raw_capacity(n: usize) -> usize {
-    n + n / 3
+    n.checked_add(n / 3)
+        .expect("overflow while converting to raw capacity")
```
hawkw, line 3224: similarly, it would be kind of nice (but again, not all that important) if this panic message included the capacity that overflowed:
```suggestion
    match n.checked_add(n / 3) {
        Some(n) => n,
        None => panic!("requested capacity {n} too large: overflow while converting to raw capacity"),
    }
```

hawkw (approved):
this looks good to me overall, thanks for the fix!

i had some minor suggestions for improving the panic messages so that the requested capacity is displayed to the user. what do you think?

src/header/map.rs
HeeillWang: reflected!

WhyNotHugo:
Function docstring should clarify this function panics:

```rs
/// # Panics
/// 
/// Requested capacity too large: would overflow `usize`.
```

HeeillWang:
> Function docstring should clarify this function panics:
> 
> ```rust
> /// # Panics
> /// 
> /// Requested capacity too large: would overflow `usize`.
> ```

reflected!

src/header/map.rs
```diff
@@ -3218,7 +3228,13 @@ fn usable_capacity(cap: usize) -> usize {
...
 #[inline]
 fn to_raw_capacity(n: usize) -> usize {
-    n + n / 3
+    match n.checked_add(n / 3) {
+        Some(n) => n,
+        None => panic!(
+            "requested capacity {} too large: overflow while converting to raw capacity",
+            n
+        ),
+    }
```
WhyNotHugo, line 3237: TBH, this could just be:

```rust
n.checked_add(n / 3).expect("requested capacity must fit usize without overflow")
```

The only downside if that the requested capacity is not shown in the error, so maybe you'll want to hear what the maintainers think before applying any changes.

(the same applies to the above `match` that basically hand-rolls an `expect`.

seanmonstar (approved):

Issue #434, opened 2020-08-11
serde support for http::Uri


jplatte:
It would be nice if `http::Uri` implemented `Deserialize` and `Serialize` under a `serde` feature flag. Would a PR be welcome for this?

jplatte:
Ping @seanmonstar

stoically:
Just in case someone ends up here and is looking for a simple workaround:

```rust
#[derive(Debug, Serialize, Deserialize)]
pub struct Uri {
    #[serde(with = "uri_serde")]
    pub uri: http::Uri,
}

mod uri_serde {
    use http::uri::InvalidUri;
    use serde::{de::Error as _, Deserialize, Deserializer, Serializer};

    pub fn deserialize<'de, D>(deserializer: D) -> Result<http::Uri, D::Error>
    where
        D: Deserializer<'de>,
    {
        let string = String::deserialize(deserializer)?;
        let uri = string.parse().map_err(|err: InvalidUri| D::Error::custom(err.to_string()))?;

        Ok(uri)
    }

    pub fn serialize<S>(uri: &http::Uri, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: Serializer,
    {
        serializer.serialize_str(&uri.to_string())
    }
}
```

tesaguri:
You can use [`http_serde::uri`].

[`http_serde::uri`]: <https://docs.rs/http-serde/1.0.3/http_serde/uri/index.html>

akiradeveloper:
What if we want `Serialize` and `Deserialize` for `HashMap<u64, Uri>`? I think the http-serde doesn't work in this case but we need a wrapper around `Uri`. This way, the application code will have a lot of wrap and unwrap, which is a bit too noisy.

SInce url crates seems has a feature gate to enable serde for the type, what is the particular reason for this crate not to provide the same feature?

andrewtoth:
I wrote a crate that provides `Serialize` and `Deserialize` for `Uri` as well as the other types from this crate. It also allows wrapping in `Option`, `HashMap`, and other `std::collection` types.
https://crates.io/crates/http-serde-ext

Issue #397, opened 2020-02-18
Add Serde support for Request and Response


KaiserKarel:
I need to log requests and responses using a structured logger. Having serde behind a feature would help a lot with that. I think these structs are quite fundamental datatypes, and that more users would probably want to store/log them somehow.

chapa:
Have you tried this crate : https://crates.io/crates/http-serde ?

(I don't, I was just passing by and I saw ~light~ your issue :blush:)

andrewtoth:
This crate lets you serialize and deserialize Request and Response types, as well as the rest of them.
https://crates.io/crates/http-serde-ext

PR #631, opened 2023-10-17
Add optional serde support for top level types

andrewtoth:
This adds `serde::Serialize` and `serde::Deserialize` trait implementations for all types re-exported at the root level, except for `Extensions`, `Result`, and `Error`. `serde` is added as an optional dependency behind the feature flag `serde1`.

These are `HeaderMap, HeaderName, HeaderValue, Method, Request, Response, StatusCode, Uri, Version, request::Parts, response::Parts`. Note for `Request` and `Response` it will fail serialization if `extensions` is non-empty, and will deserialize to a `Extensions::default()`.

Current workarounds are to use the `http_serde` crate, but that requires annotating `#[serde(with)]` and doesn't support using `Option` or other container types. Suggested alternatives are [quite complex](https://stackoverflow.com/questions/76143129/rust-serde-http-serialize-optionhttpmethod).

There was a previous attempt just for `StatusCode` in https://github.com/hyperium/http/pull/274 which seems to have gone stale.

Implements https://github.com/hyperium/http/issues/397, https://github.com/hyperium/http/issues/273, https://github.com/hyperium/http/issues/434

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,7 @@ jobs:
 
       - name: Test
         run: cargo test
+        run: cargo test --all-features
 
       - name: Test all benches
         if: matrix.benches
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,10 @@ exclude = [
 bytes = "1"
 fnv = "1.0.5"
 itoa = "1"
+serde = { version = "1.0", optional = true, features = ["derive"] }
+
+[features]
+serde1 = ["serde"]
 
 [dev-dependencies]
 indexmap = "<=1.8"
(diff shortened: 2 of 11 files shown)


seanmonstar:
Reading the old issue, it's been a while since I've thought about this. With some of the ecosystem recently worried about serde (not that I particularly agree with the worry itself), I find myself leaning back towards not including it directly in `http`. 🙈

Other maintainers could disagree. But it seems sufficiently solvable with a separate crate.

andrewtoth:
A fair point. Thank you for sharing your thoughts on this.  I've created a crate to solve this using the `serde(with = "...")` attributes in case anyone else comes here looking for a solution.
https://crates.io/crates/http-serde-ext

PR #634, opened 2023-11-09
feat: implement Clone for Request, Response, Extensions

### LINKED ISSUE #395: Update `extensions` to be cloneable

seanmonstar:
A breaking change, requiring all extensions to `impl Clone`.

Closes #395

### DIFF
--- a/src/extensions.rs
+++ b/src/extensions.rs
@@ -3,7 +3,7 @@ use std::collections::HashMap;
 use std::fmt;
 use std::hash::{BuildHasherDefault, Hasher};
 
-type AnyMap = HashMap<TypeId, Box<dyn Any + Send + Sync>, BuildHasherDefault<IdHasher>>;
+type AnyMap = HashMap<TypeId, Box<dyn AnyClone + Send + Sync>, BuildHasherDefault<IdHasher>>;
 
 // With TypeIds as keys, there's no need to hash them. They are already hashes
 // themselves, coming from the compiler. The IdHasher just holds the u64 of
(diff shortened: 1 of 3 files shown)


LucioFranco (approved):

src/extensions.rs
```diff
@@ -59,12 +59,12 @@ impl Extensions {
     /// assert!(ext.insert(4u8).is_none());
     /// assert_eq!(ext.insert(9i32), Some(5i32));
     /// ```
-    pub fn insert<T: Send + Sync + 'static>(&mut self, val: T) -> Option<T> {
+    pub fn insert<T: Clone + Send + Sync + 'static>(&mut self, val: T) -> Option<T> {
```
kvcache, line 62: This breaks tonic interceptors that attach extensions with well-defined ownership relying on `&mut self`.

My tonic interceptors that attach internal metrics and internal metadata now have to have something like `Arc<Mutex<Option<MyExtension>>>`.
Before:
```rust
let mut extension = request.extensions_mut().remove().expect("extension was set");
```

After:
```rust
let extension: Arc<Mutex<Option<ReusableOwned<Detailer>>>> = request
    .extensions_mut()
    .remove()
    .unwrap_or_default();
let extension = std::mem::take(&mut *extension.lock().expect("one locker"));
let mut extension = extension.expect("extension was set");
```

seanmonstar: Yes, this is expected. It's a trade-off, and we settled on merging this.

PR #637, opened 2023-11-10
feat: add default std feature

seanmonstar:
cc #563


algesten (approved):
Looks good!

LucioFranco (approved):

seanmonstar:
I admit I have no clue why the compilation fails so completely...

src/lib.rs
```diff
@@ -160,6 +160,11 @@
 
 #![deny(warnings, missing_docs, missing_debug_implementations)]
 
+//#![cfg_attr(not(feature = "std"), no_std)]
+#![cfg(not(feature = "std"))]
```
seanmonstar, line 164: Oh, that `!` was disabling the entire crate. lul.

Issue #639, opened 2023-11-12
HeaderMap error: Undefined Behavior: trying to retag from _  for Unique permission


hjr3:
In https://github.com/hyperium/hyper/pull/3375 we ran into the following error:

```
error: Undefined Behavior: trying to retag from <1400070> for Unique permission at alloc405811[0x28], but that tag only grants SharedReadOnly permission for this location
    --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/http-0.2.10/src/header/map.rs:2111:35
     |
2111 |         let entry = unsafe { &mut (*self.map).entries[self.entry] };
     |                                   ^^^^^^^^^^^^^^^^^^^
     |                                   |
     |                                   trying to retag from <1400070> for Unique permission at alloc405811[0x28], but that tag only grants SharedReadOnly permission for this location
     |                                   this error occurs as part of retag at alloc405811[0x28..0x40]
     |
     = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
     = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
help: <1400070> was created by a SharedReadOnly retag at offsets [0x0..0x60]
    --> src/proto/h1/role.rs:1524:26
     |
```

It was observed that this error is not occurring in `v0.2.9`.

I reproduced the error with this test in `test/header_map.rs`:

```rust
#[test]
fn repro_retag_error() {
    let mut headers = HeaderMap::new();
    headers.insert(
        HeaderName::from_static("chunky-trailer"),
        HeaderValue::from_static("header data"),
    );

    let _foo = &headers.iter().next();
}
```

I did a `git bisect` and identified the error:

```
$ git bisect start
$ git bisect good v0.2.9
$ git bisect bad v0.2.10
$ git bisect run sh -c 'RUSTFLAGS=-Adropping_copy_types MIRIFLAGS="-Zmiri-disable-isolation" cargo +nightly miri test chunked_with_title_case_headers'

...

78e3d37563b0fe83886abffdcd8569117d45fee9 is the first bad commit
commit 78e3d37563b0fe83886abffdcd8569117d45fee9
Author: Discord9 <discord9@163.com>
Date:   Thu Jul 13 18:29:29 2023 +0800

    make miri happy

 src/header/map.rs | 8 ++++----
 1 file changed, 4 insertions(+), 4 deletions(-)
bisect found first bad commit
[...]

PR #640, opened 2023-11-13
Fix miri error

hjr3:
Fixes https://github.com/hyperium/http/issues/639


seanmonstar:
Seems like the header fuzz test was perhaps timing it out or something. I made #641 just to try to get this fixed ASAP.

seanmonstar:
Hm, nope, there's still MIRI failures. I've opened #642 that should no longer reverts, but fixes the other problem that was noticed.

seanmonstar:
That fixed it, so I'll close. Thanks for the help!

PR #641, opened 2023-11-13
Fix MIRI test

seanmonstar:
Adds onto #640


seanmonstar:
Hm, nope, there's still MIRI failures. I've opened #642 that should no longer reverts, but fixes the other problem that was noticed.

PR #642, opened 2023-11-13
fix MIRI error in header::Iter

seanmonstar:


hjr3:
Should we still add miri to the CI so we do not regress this?

seanmonstar:
D'oh, I thought I include that in this PR. Yes, we should add that. (I was testing locally.)

hjr3:
I will put a PR up.

PR #676, opened 2024-02-16
refactor(lib): move request and response trait bound assertion to test

tottoto:
Refactors trait bound assertions.


seanmonstar:
Is there a benefit to this move?

tottoto:
Since what `_assert_types` doing is only for tests, I would suppose moving it to test would be easier to read for developers.

seanmonstar (approved):
Okay!

PR #677, opened 2024-02-16
refactor(lib): apply small refactoring

tottoto:
Applies miscellaneous refactoring.


src/request.rs
```diff
@@ -165,6 +165,7 @@ pub struct Request<T> {
 /// The HTTP request head consists of a method, uri, version, and a set of
 /// header fields.
 #[derive(Clone)]
+#[non_exhaustive]
```
seanmonstar, line 168: Would you mind making `#[non_exhaustive]` a different PR?

seanmonstar (commented):
Very nice!

src/request.rs
tottoto: Moved this commit to #678.

seanmonstar (approved):

PR #680, opened 2024-02-21
Allow using `as_u16` in const contexts

WhyNotHugo:


src/status.rs
```diff
@@ -116,8 +116,8 @@ impl StatusCode {
     /// assert_eq!(status.as_u16(), 200);
     /// ```
     #[inline]
-    pub fn as_u16(&self) -> u16 {
-        (*self).into()
+    pub const fn as_u16(&self) -> u16 {
+        (*self).0.get()
```
WhyNotHugo, line 120: Hint for reviewers: `into()` did exactly this under the hood.

seanmonstar (approved):

PR #682, opened 2024-03-01
feat: add `HeaderMap::try_` methods to handle capacity overflow

seanmonstar:
This continues #617. Merge conflicts were resolved, and a case of a breaking change was resolved since I forgot `try_entry()` needed to change return types before we release 1.0. _Sigh_.

### DIFF
--- a/src/header/map.rs
+++ b/src/header/map.rs
@@ -213,6 +213,11 @@ pub struct ValueDrain<'a, T> {
     lt: PhantomData<&'a mut HeaderMap<T>>,
 }
 
+/// Error returned when max capacity of `HeaderMap` is exceeded
+pub struct MaxSizeReached {
+    _priv: (),
+}
+
 /// Tracks the value iterator state
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
 enum Cursor {
@@ -441,7 +446,7 @@ impl HeaderMap {
     /// assert_eq!(0, map.capacity());
     /// ```
     pub fn new() -> Self {
-        HeaderMap::with_capacity(0)
+        HeaderMap::try_with_capacity(0).unwrap()
     }
 }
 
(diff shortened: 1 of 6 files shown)


src/header/map.rs
```diff
@@ -1068,22 +1135,33 @@ impl<T> HeaderMap<T> {
...
         K: AsHeaderName,
     {
-        key.try_entry(self)
+        key.try_entry(self).map_err(|err| match err {
+            as_header_name::TryEntryError::InvalidHeaderName(e) => e,
+            as_header_name::TryEntryError::MaxSizeReached(_e) => {
+                // Unfortunately, we cannot change the return type of this
+                // method, so the max size reached error needs to be converted
+                // into an InvalidHeaderName. Yay.
+                InvalidHeaderName::new()
```
glebpom, line 1151: Would it make sense to mark `try_entry` as deprecated?

seanmonstar: And introduce a new method? With what name?

src/header/map.rs
```diff
@@ -1068,22 +1135,33 @@ impl<T> HeaderMap<T> {
     /// valid `HeaderName`s to passed as the key (such as `String`). If they
     /// do not parse as a valid `HeaderName`, this returns an
     /// `InvalidHeaderName` error.
+    ///
+    /// It may return `MaxSizeReached` error if size exceeds the maximum
+    /// `HeaderMap` capacity
```
glebpom, line 1140: This comment is inaccurate; it actually returns `InvalidHeaderName`

src/header/map.rs
glebpom: Would `try_entry2` work, along with renaming private `try_entry2` to something else?

seanmonstar: It's a possibility. I'd likely wait to see if that is something that is requested much.

PR #688, opened 2024-03-13
Fix `html_root_url`

LukeMathWalker:


LukeMathWalker:
Although an actual fix requires setting it to `0.2.13` and cutting a new release.  
What is the purpose of setting the attribute in the first place?

seanmonstar:
It's legacy. Docs now have a good default, we can remove the attribute.

LukeMathWalker:
Done in [0b1d17b](https://github.com/hyperium/http/pull/688/commits/0b1d17b052688bdf9e91c57dfa128f6fb21affc2) 👍🏻

seanmonstar (approved):

PR #689, opened 2024-03-16
feat: add strict version of uri parsing

dswij:
This PR adds a strict version of uri parsing to disallow `{`, `}` and `"`.

See https://github.com/hyperium/hyper/issues/3594

### DIFF
--- a/src/uri/mod.rs
+++ b/src/uri/mod.rs
@@ -339,6 +339,57 @@ impl Uri {
         parse_full(s)
     }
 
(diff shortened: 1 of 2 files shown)


seanmonstar:
I don't think this is a way to go. Part of the point of a specific type is knowing that once parsed, it matches a certain set of characters. Having different constructors could mean two different instances could have separate character sets.

This differs in practice from how `HeaderName::from_lowercase()`, because all the constructors still force to lowercase, that one constructor just adds an additional validation, instead of coercing to the same character set.

PR #692, opened 2024-03-20
Rename UNPROCESSABLE_ENTITY to UNPROCESSABLE_CONTENT

### LINKED ISSUE #664: Rename 422 from "Unprocessable Entity" to "Unprocessable Content"
## Present State

Currently: [status.rs lines 456 to 458](https://github.com/hyperium/http/blob/bda93204b3da1a776cf471ed39e8e374cec652e7/src/status.rs#L456):

```rust
/// 422 Unprocessable Entity
/// [[RFC4918](https://tools.ietf.org/html/rfc4918)]
(422, UNPROCESSABLE_ENTITY, "Unprocessable Entity");
```

## Recommendation

Use this instead:

```rust
/// 422 Unprocessable Content
/// [[RFC9110](https://datatracker.ietf.org/doc/html/rfc9110#section-15.5.21)]
(422, UNPROCESSABLE_CONTENT, "Unprocessable Content")
```

## Relevant RFCs

Both of the following RFCs define 422 but with a one-word difference:

- [RFC 4918](https://datatracker.ietf.org/doc/html/rfc4918), "Proposed Standard", June 2007, "Unprocessable **Entity**"
- [RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110), "Internet Standard", June 2022, "Unprocessable **Content**"

**Potential Confusion**: Even though RFC 9110 it is newer, it does not obsolete RFC 4918 (directly nor transitively).

### Arguments in Favor of RFC 9110

1. RFC 9110 is titled "HTTP Semantics" which is a direct fit for this library, whereas RFC 4918 is scoped to WebDAV.

2. RFC 9110 is an [Internet Standard](https://en.wikipedia.org/wiki/Internet_Standard) while RFC 4918 is only a [Proposed Standard](https://en.wikipedia.org/wiki/Internet_Standard#Proposed_Standard).

3. [MDN uses RFC 9110 for the 422 status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422)

4. RFC 9110 is newer (2022 vs. 2007)

## RFC 9110: 422: Unprocessable Content

> [15.5.21.](https://httpwg.org/specs/rfc9110.html#rfc.section.15.5.21) [422 Unprocessable Content](https://httpwg.org/specs/rfc9110.html#status.422)
>
> The 422 (Unprocessable Content) status code indicates that the server understands the content type of the request content (hence a [415 (Unsupported Media Type)](https://httpwg.org/specs/rfc9110.html#status.415) status code is inappropriate), and the syntax of the request content is correct, but it was unable to process the contained instructions. For example, this status code can be sent if an XML request content contains well-formed (i.e., syntactically correct), but semantically erroneous XML instructions.

## References

- https://github.com/mdn/content/pull/24866
- https://httpwg.org/specs/rfc9110.html#status.422
- https://stackoverflow.com/a/52098667

Pi-Cla:
Fix #664


seanmonstar:
I believe this would be a breaking change, since an existing constant would no longer exist.

@obi1kenobi 👋  curious, is this something that the semver job should be able to notice?

seanmonstar:
I think another PR exists which modifies the macro to allow "aliases".

obi1kenobi:
> I believe this would be a breaking change, since an existing constant would no longer exist.
> 
> @obi1kenobi 👋 curious, is this something that the semver job should be able to notice?

Thanks for the ping, I appreciate it!

We currently have the equivalent lint for free (module-level) constants, but don't have the lint for constants inside a type's `impl` block yet: https://github.com/obi1kenobi/cargo-semver-checks/issues/366#issuecomment-1662820577

It's very easy to add, and we have a few contributors who are actively adding lints right now, so we can prioritize it without any trouble. It'll be in the next release for sure.

seanmonstar:
Ya no worries, I know it's neverending work. Thanks so much!

obi1kenobi:
Thank you for using it, and for the supportive feedback! It really helps me prioritize things.

If there are other things that would be useful, I'd love to hear about them! Either a GitHub issue or a 15min chat (and I can file issues as needed) would work. Ultimately a lot of things are funding-limited, but it helps me know what to say users want when I apply for OSS grants!

obi1kenobi:
Just merged the lint that will catch issues like this going forward: https://github.com/obi1kenobi/cargo-semver-checks/pull/714

It'll be part of the next `cargo-semver-checks` release, planned for a day two after Rust 1.78 so ~first week of May.

One fewer breaking change that can sneak unnoticed into our crates!

obi1kenobi:
I ended up cutting a new release earlier than originally planned: https://github.com/obi1kenobi/cargo-semver-checks/releases/tag/v0.31.0

The new `inherent_associated_pub_const_missing` lint will catch cases like this going forward. Thanks again for your patience and support, and for maintaining so many awesome crates! ❤️

mina86:
> I think another PR exists which modifies the macro to allow "aliases".

https://github.com/hyperium/http/pull/588

PR #694, opened 2024-04-09
simplified code

wandercn:
remove generic T

### DIFF
--- a/src/uri/authority.rs
+++ b/src/uri/authority.rs
@@ -235,7 +235,7 @@ impl Authority {
     ///
     /// assert!(authority.port().is_none());
     /// ```
-    pub fn port(&self) -> Option<Port<&str>> {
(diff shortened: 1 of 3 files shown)


tesaguri:
This is a breaking change and I don't think it is going to be accepted, although I'm not the maintainer of the project.

Alternatively, you could add a default type to the `T` parameter, which is a non-breaking change, but in that case, the default parameter wouldn't be able to `&str` because that would require a lifetime parameter, and `String` wouldn't be appropriate as the default type for many use cases either due to a performance reason.

So, what is your motivation behind the change? If you just don't want to write type parameters in your code, you can define a type alias like `type Port = http::uri::Port<String>;` in your code.

seanmonstar:
Yea, this is breaking, we cannot do this. I'm going to close, but if you have a problem you're trying to solve, you can open a discussion and say more 👍

PR #699, opened 2024-05-22
feat(status-code): `impl Deref for StatusCode` to read the inner value

sivizius:
With this PR, it is possible to access the inner `NonzeroU16` value without modifying it. This avoids `NonZeroU16::try_from(status_code.as_u16()).unwrap()`, which introduces a panic, which will be optimised away by the compiler, but e.g. clippy might complain about a missing `# Panic`-section in the documentation. Due to `impl Copy for NonZeroU16`, one can simply convert `StatusCode` to `NonZeroU16`. `DerefMut` was not implemented, because this would make the range-check useless.

### DIFF
--- a/src/status.rs
+++ b/src/status.rs
@@ -18,6 +18,7 @@ use std::convert::TryFrom;
 use std::error::Error;
 use std::fmt;
 use std::num::NonZeroU16;
+use std::ops::Deref;
 use std::str::FromStr;
 
 /// An HTTP status code (`status-code` in RFC 9110 et al.).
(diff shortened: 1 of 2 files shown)


seanmonstar:
I would not merge this. This exposes the inner representation of the type, which would prevent ever changing it.

seanmonstar:
> Will this ever change?

Possibly not, but it's good practice for libraries to keep internals private to allow for that without breaking changes. If the representation did change, we could still cheaply produce a `u16` in the `as_u16()` method.

The current solution, `.as_u16().try_into().unwrap()`, looks fine to me.

PR #703, opened 2024-06-15
refactor(lib): use `#[non_exhaustive]`

SandaruKasa:
This seems to be supported starting with 1.40 (the MSRV is 1.48) and accomplishes the exact same functionality.

Got caught by clippy btw.


SandaruKasa:
I would probably add some tests for this, but I don't think there's an easy way to add fails-to-compile tests without a 3rd party framework.

tottoto:
Duplicate of #678.

PR #705, opened 2024-06-21
Improve documentation for the `insert` method

asudox:


src/extensions.rs
```diff
@@ -48,7 +48,7 @@ impl Extensions {
     /// Insert a type into this `Extensions`.
     ///
     /// If a extension of this type already existed, it will
-    /// be returned.
+    /// be returned and replaced with the new one.
```
seanmonstar, line 51:
```suggestion
    /// be returned and replaced with the new one.
```

seanmonstar (approved):
Thank you!

PR #713, opened 2024-08-07
Allow characters `#`, `$`, `%`, `&`, `'` in HTTP method

franfastly:
This PR adds those missing characters to `http::method::extension::METHOD_CHARS` and updates its comment to reference RFC 9110.


seanmonstar (approved):
Thanks for noticing, and fixing this!

PR #715, opened 2024-08-29
feat(path-and-query): allow utf8 char, not rfc 3986 compliant, in path and query

joelwurtz:
This is a follow up of a PR in https://github.com/seanmonstar/httparse/pull/178 

This is mainly to allow incoming request that may have utf8 char in their path or query as some browsers send them raw instead of doing a percent encode (see the details here : https://github.com/actix/actix-web/issues/3102#issuecomment-1676454659 for a list of browser doing that)

This patch allow to correctly build a request object based on this behavior

I made it simple for the moment as i don't know how you want to handle this, in a feature flag ? through an option ? neither of them ?

Also i don't know on which version this should be based, it seems that 0.2 still receive some support should it be backported to this version also ?

### DIFF
--- a/src/uri/path.rs
+++ b/src/uri/path.rs
@@ -50,7 +50,7 @@ impl PathAndQuery {
                     0x40..=0x5F |
                     0x61..=0x7A |
                     0x7C |
-                    0x7E => {}
+                    0x7E..=0xFF => {}
 
                     // These are code points that are supposed to be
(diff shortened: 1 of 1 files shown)


seanmonstar (approved):
Thanks for submitting this! After talking with a few other people, this seems like the right way to go.

joelwurtz:
Thanks for merging

PR #717, opened 2024-10-11
feat: add support for HTTP QUERY method

jgresty:
The HTTP QUERY method is a means of making a safe, idempotent request that contains content. The primary motivator for this case is graphql, replacing HTTP POST for read only requests, however it is application agnostic.

This method is still in draft but stable enough where it is unlikely to change for this implementation.

https://httpwg.org/http-extensions/draft-ietf-httpbis-safe-method-w-body.html

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -413,6 +413,28 @@ impl Request<()> {
     {
         Builder::new().method(Method::TRACE).uri(uri)
     }
+
+    /// Creates a new `Builder` initialized with a QUERY method and the given URI.
+    ///
+    /// This method returns an instance of `Builder` which can be used to
+    /// create a `Request`.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use http::*;
+    ///
+    /// let request = Request::query("https://www.rust-lang.org/")
+    ///     .body(())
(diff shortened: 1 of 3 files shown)


gdubicki:
I think you should add support for the `accept-query` header too, @jgresty, see and feel free to take the code from my draft PR: https://github.com/hyperium/http/pull/745

seanmonstar:
I'd prefer to leave off the header for now. I think the method is probably a fine addition, but I'd like to wait for real world usage of the header to possibly grow (or for servers to not care to send it), before deciding to add to the library.

src/request.rs
```diff
@@ -413,6 +413,28 @@ impl Request<()> {
...
+    /// # Example
+    ///
+    /// ```
+    /// # use http::*;
+    ///
+    /// let request = Request::query("https://www.rust-lang.org/")
+    ///     .body(())
+    ///     .unwrap();
+    /// ```
+    pub fn query<T>(uri: T) -> Builder
```
seanmonstar, line 431: Hm, this is my opinion here: the method name here could also be used to refer to the a query portion of the URI. I think we could leave this off.

PR #720, opened 2024-10-25
feat(wasi): add `wasi` crate support

rvolosatovs:
This adds some very basic and common conversions to and from [`wasi:http` types](https://github.com/WebAssembly/wasi-http/blob/6a86347b54b10d1c019caf15b229ac0bd918403c/wit/types.wit) using [generated bindings from `wasi` crate](https://docs.rs/wasi/latest/wasi/http/types/index.html).

This opt-in feature is gated behind `wasi` feature flag.

I did not write a test case for the header conversion to avoid pulling in a runtime at this point.

Refs:
- https://github.com/bytecodealliance/wasi-rs/pull/90
- https://github.com/wasmCloud/wasmCloud/pull/3444


seanmonstar:
I don't think we'd add a public dependency here. Adding _any_ public dependency is a risky move, and wasi isn't stable yet, so more so. Once a dependency is publicly exposed, it's semver becomes ours. We don't want to have to release a v2 to upgrade a dependency.

I clicked around the links you mentioned, and it seems like support already exists in another crate.

rvolosatovs:
> I don't think we'd add a public dependency here. Adding _any_ public dependency is a risky move, and wasi isn't stable yet, so more so. Once a dependency is publicly exposed, it's semver becomes ours. We don't want to have to release a v2 to upgrade a dependency.
> 
> I clicked around the links you mentioned, and it seems like support already exists in another crate.

fair enough

PR #729, opened 2024-11-26
Generalize HeaderMap's TryFrom impl from HashMap, to allow other hashing algorithms

### LINKED ISSUE #726: HeaderMap's `TryFrom<HashMap>` is not generic over the hashing algorithm used
`HeaderMap<T>` implements [`TryFrom<&'a HashMap<K, T>>`](https://docs.rs/http/1.1.0/http/header/struct.HeaderMap.html#impl-TryFrom%3C%26HashMap%3CK,+V%3E%3E-for-HeaderMap%3CT%3E). However, this only covers hashmaps created with the default `S = RandomState` hashing algorithm / state.
If you've gathered some headers from elsewhere (in my case: in a Python library with PyO3) and you want to turn those into a HeaderMap for use in a HTTP client, there is no need to use a HashDoS-resistant hashing function, so being able to use a hashmap with a different hashing function would be great.

As far as I can tell, there is no reason why the current implementation could not work for other hashmaps. The only thing that needs to change is exposing `S` parameter as well in the trait implementation bounds.

Qqwy:
Instead of `impl TryFrom<&'a HashMap<K, V>> for HeaderMap<T>` this PR changes this to `impl TryFrom<&'a HashMap<K, V, S>> for HeaderMap<T>`.
No change in behaviour, it only makes the input type more general so it allows passing hashmaps with other hashing functions as well.

This change is fully backwards-compatible because of the no-orphans rule.

Fixes #726

### DIFF
--- a/src/header/map.rs
+++ b/src/header/map.rs
@@ -2020,7 +2020,7 @@ impl<T> FromIterator<(HeaderName, T)> for HeaderMap<T> {
 /// let headers: HeaderMap = (&map).try_into().expect("valid headers");
 /// assert_eq!(headers["X-Custom-Header"], "my value");
 /// ```
-impl<'a, K, V, T> TryFrom<&'a HashMap<K, V>> for HeaderMap<T>
+impl<'a, K, V, S, T> TryFrom<&'a HashMap<K, V, S>> for HeaderMap<T>
 where
     K: Eq + Hash,
     HeaderName: TryFrom<&'a K>,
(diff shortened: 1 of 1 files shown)


seanmonstar (approved):
Thanks so much!

Qqwy:
@seanmonstar When will you publish a new version that contains this change?

seanmonstar:
Let's aim for Monday (prefer not to do over a weekend.)

PR #730, opened 2024-11-28
#727 Use TryInto Traitbounds for Builder Methods

### LINKED ISSUE #727: Prefer TryInto over TryFrom in builders
I have been looking at the API for this crate as someone relatively new to Rust.
Amongst the code, I noticed this method deceleration for the requests::Builder struct
```rust
pub fn method<T>(self, method: T) -> Builder
where
    Method: TryFrom<T>,
    <Method as TryFrom<T>>::Error: Into<Error>,
```
Initially I was confused as to the reasoning that a constraint on `Method` was defined here.
Since `Method` seems to be unused in the function signature.
After looking deeper I realize the point of this constraint is to limit `T` to values that can be converted to `Method::try_from`. 
Is there any reason that the where clause doesn't use the reciprocal trait `TryInto` instead?
For example,
```rust
pub fn method<T>(self, method: T) -> Builder
where
    T TryInto<Method>,
    <T as TryInto<Method>>::Error: Into<Error>,
```
Idk about others but this would have been clearer to me.
I was curious if there was a reason as to why the original was chosen over this second solution, which I think is more clear(imo).

SmolPatches:
Closes #727 
Changes to use TryInto Trait Bounds for Builder methods.
Let me know if any additional edits need to be made!
Happy Turkeyday!

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -53,7 +53,7 @@
 //! ```
 
 use std::any::Any;
-use std::convert::TryFrom;
+use std::convert::TryInto;
 use std::fmt;
 
 use crate::header::{HeaderMap, HeaderName, HeaderValue};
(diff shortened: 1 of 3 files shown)


seanmonstar (approved):
Thanks so much!

PR #731, opened 2024-11-28
Bump MSRV to 1.70

marmeladema:
This project follows Tokio MSRV which is now at 1.70.
Additionally, this allows to use panic in constant to simplify some code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ keywords = ["http"]
 categories = ["web-programming"]
 edition = "2018"
 # When updating this value, don't forget to also adjust the GitHub Actions config.
-rust-version = "1.49.0"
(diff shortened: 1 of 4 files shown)


seanmonstar:
Hm, that comment in the README is likely wrong, this more closely follows hyper's policy: https://github.com/hyperium/hyper/blob/master/docs/MSRV.md

Which I'm not so sure I'd want to increase it for better panic messages, as Debian stable still has 1.63.

marmeladema:
Thank you for your inputs.
Bumping to 1.63 would still be enough to get const panics.
I will update the proposal to that version but if that's still not appropriate, feel free to close this.

seanmonstar:
Closing, but if you want to submit a bump to 1.63, we could merge that!

Issue #551, opened 2022-05-10
Move `uri` module into extra crate


Stupremee:
Hello,

I'm currently in need of a correct `Uri` parser, and I believe the best URI implementation is the one in this crate.
However, I don't want to add the whole `http` crate as a dependency.
So, it would be very helpful if there was an additional crate just for URI parsing.

LucioFranco:
Is there a reason you don't want to just bring in the whole http crate? The dependency is already quite light. I don't think moving things to their own crate makes anyone's life easier.

Stupremee:
The reason is that `http` does not have support for `no_std` environments.

LucioFranco:
Maybe a better goal would be to allow http to support no_std similar to how `bytes` has no_std support. I don't think the correct solution is to create another crate.

Stupremee:
Sure, that would work as well!

Erik1000:
+1 for this

olanod:
My use case is an HTTP client running in embedded hardware, hope this PR(https://github.com/hyperium/http/pull/563) can get some feedback :) Ideally it would be nice if `http` works without `alloc` but that seemed like a mayor refactor and possibly breaking change.

ivila:
@Stupremee If you just need a Uri parser, what about the [url crate from Servo](https://crates.io/crates/url), it supports `no_std` starts from `2.5.2`.

mspiegel:
I have a fork that has no_std support. It is available at https://github.com/hushmesh/http. Should I submit a PR to merge it back into this repository?

The following changes are made: 1) struct Extensions uses hashbrown::HashMap instead of std::collections::HashMap. 2) The adaptive hashing algorithm in the HeaderMap uses std::collections::hash_map::RandomState when std is enabled (same as current behavior) and uses foldhash::quality::RandomState when std is disabled. 3) The MSRV is bumped to 1.81.

I think the bump of the MSRV is the most significant change. If 1.81 is not old enough, I could go ahead and try the technique where a "prelude" namespace is declared and "prelude" maps to "std::" when the std feature is enabled or "core::" when the std feature is disabled. That usually allows for a lower MSRV but only for the std feature. 1.81 would be the MSRV when std is disabled (something that cannot be expressed in Cargo.toml)

[...]

PR #732, opened 2024-12-01
wip: no_std

brandonros:
part of https://github.com/hyperium/http/issues/551

I got stuck on `Error` and `RandomState`, but other than that this should be close?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ exclude = [
 ]
 
 [features]
-default = ["std"]
+default = []
 std = []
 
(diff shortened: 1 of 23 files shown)


Cargo.toml
```diff
@@ -39,6 +39,7 @@ std = []
 [dependencies]
 bytes = "1"
 fnv = "1.0.5"
+hashbrown = "0.12.0"
```
brandonros, line 42: not sure how to put this behind a no_std flag (replaces std::HashMap)

PR #739, opened 2024-12-28
Allow non-percent-encoded spaces in path

WhyNotHugo:
The space character is not a reserved character. While it MAY be used in percent-encoded form, this is not mandatory and it is acceptable to use raw spaces in URL paths.

Add Space (0x20) to the list of characters that don't need to be percent-encoded.

### DIFF
--- a/src/uri/path.rs
+++ b/src/uri/path.rs
@@ -44,7 +44,7 @@ impl PathAndQuery {
                     // percent-encoded in the path. If it should have been
                     // percent-encoded, then error.
                     #[rustfmt::skip]
-                    0x21 |
+                    0x20..=0x21 |
                     0x24..=0x3B |
                     0x3D |
                     0x40..=0x5F |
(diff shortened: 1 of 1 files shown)


seanmonstar:
I don't see any feasible way this could work, at least for HTTP/1. Parsers assume (because the spec defines this) that the request line is 3 tokens separate by spaces. A parser would be upset to see `GET /hello world HTTP/1.1`.

Note [RFC 9112 3.2](https://httpwg.org/specs/rfc9112.html#rfc.section.3.2):

> No whitespace is allowed in the request-target. Unfortunately, some user agents fail to properly encode or exclude whitespace found in hypertext references, resulting in those disallowed characters being sent as the request-target in a malformed request-line.

WhyNotHugo:
You’re right, this won’t work. Thanks for pointing me in the right direction.

Aside from a space character, is there any other non-reserved URL character that needs to be escaped?

WhyNotHugo:
recipients MAY instead parse on whitespace-delimited word boundaries and, aside from the CRLF terminator, treat any form of whitespace as the SP separator while ignoring preceding or trailing whitespace; such whitespace includes one or more of the following octets: SP, HTAB, VT (%x0B), FF (%x0C), or bare CR.

WhyNotHugo:
I'm working on WebDAV, where URLs are sometimes specified in the request header, and sometimes inside an XML in the body. I've been using the same Uri type for both, but it seems that rules for percent-encoding the URIs inside the XML body are more relaxed.

Should I just use a separate type for those?

PR #740, opened 2025-01-02
no_std support.

### LINKED ISSUE #551: Move `uri` module into extra crate

sunfishcode:
This PR is similar to other no-std PRs, however it takes the approach of using the new [`core::error`] module in Rust 1.81. This means that no-std mode has an MSRV of Rust 1.81, while the existing MSRV of 1.49 is still supported for existing users, as suggested [here].

This PR also preserves semver compatibility, and avoids adding any new dependencies or required features for existing users. And it avoids modifying the tests and benchmark sources, as those don't need to be no-std. And it avoids making any unrelated changes.

And, it adds CI coverage and README.md documentation.

[here]: https://github.com/hyperium/http/pull/563#issuecomment-2257330014
[`core::error`]: https://doc.rust-lang.org/stable/core/error/index.html

Fixes #551.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,6 +88,19 @@ jobs:
       - name: Test
         run: cargo check -p http
 
+  no-std:
+    name: Check no_std
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Check
+        run: cargo check --no-default-features --features=no-std
+
   wasm:
     name: WASM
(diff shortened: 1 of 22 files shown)


ivila:
Just a suggestion, maybe you can create a compatible layer like [serde-rs](https://github.com/serde-rs/serde) do: https://github.com/serde-rs/serde/blob/master/serde/src/lib.rs#L166-L287
image

Shua-github:
Sorry to bother you, but has there been any progress on this PR?

PR #741, opened 2025-01-08
fix: HeaderMap::reserve allocates insufficient capacity

kawaemon:
Current implementation of `HeaderMap::reserve` actually reserves less capacity than requested, causing reallocation when attempted to insert the requested number of entries.

This PR adds a test, then fixes it by converting requested capacity to raw one before allocation.


seanmonstar (approved):
Nice improvement, thank you!

PR #744, opened 2025-01-12
Add QUERY method support

### LINKED ISSUE #743: Please add support for the `QUERY` HTTP method
On 2025-01-07 the [`QUERY` HTTP method, "a safe, idempotent request method that can carry request content." has reached the level of a Proposed Standard](https://datatracker.ietf.org/doc/draft-ietf-httpbis-safe-method-w-body/).

It solves the problem of making complex queries where the GET method is sometimes sufficient as it (according to the standard) lacks support for sending the query in the body, leaving only the query part of the URL which is limited. An often used workaround of using the POST method has a downside in that it's just the wrong method for this purpose as queries do not cause side-effects and this limits the retryabilty and cacheability of such queries.

Therefore It would be great to implement support for `QUERY` in this library.

If you are open to it, I am interested in providing a PR.

(Related: https://github.com/tokio-rs/axum/issues/1691)

sigmavirus24:
[draft-ietf-httpbis-safe-method-w-body-07][] was recently accepted by the HTTP WG (although so recently that it does not have an official RFC number).

The method and requests using it are meant to be idempotent and retryable.

Closes #743

[draft-ietf-httpbis-safe-method-w-body-07]: https://datatracker.ietf.org/doc/draft-ietf-httpbis-safe-method-w-body/

### DIFF
--- a/src/method.rs
+++ b/src/method.rs
@@ -60,6 +60,7 @@ enum Inner {
     Trace,
     Connect,
     Patch,
+    Query,
     // If the extension is short enough, store it inline
     ExtensionInline(InlineExtension),
     // Otherwise, allocate it
@@ -94,6 +95,9 @@ impl Method {
     /// TRACE
     pub const TRACE: Method = Method(Trace);
 
+    /// QUERY
+    pub const QUERY: Method = Method(Query);
+
     /// Converts a slice of bytes to an HTTP method.
     pub fn from_bytes(src: &[u8]) -> Result<Method, InvalidMethod> {
         match src.len() {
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Withdrawing this, a closer read of the stages of the IETF means this might still not make it to final RFC status.

gdubicki:
This PR was missing `accept-query` header support. See https://github.com/hyperium/http/pull/745 instead.

> Withdrawing this, a closer read of the stages of the IETF means this might still not make it to final RFC status.

True, but if it reached the consensus level the it should reach its target Proposed Standard status (like f.e. the PATCH method).

And many technologies have been widely implemented before they have reached their target status. For example [HTTP/3 was released in most browsers many months before it reached it](https://en.wikipedia.org/wiki/HTTP/3#History).

PR #745, opened 2025-01-12
feat: add support for HTTP QUERY method

gdubicki:
The HTTP QUERY method is a means of making a safe, idempotent request that contains content. The primary motivator for this case is graphql, replacing HTTP POST for read only requests, however it is application agnostic.

This method is still in draft but since 2025-01-07 it has reached consensus so it should be stable enough to make it unlikely to change this implementation.

https://datatracker.ietf.org/doc/draft-ietf-httpbis-safe-method-w-body/

Closes https://github.com/hyperium/http/issues/743


gdubicki:
This is an alternative, (arguably?) a bit more complete version of https://github.com/hyperium/http/pull/717. IMHO either that PR should be expanded and this one closed or this should take over.

gdubicki:
Closing in favor of #717, see [this](https://github.com/hyperium/http/pull/717#issuecomment-2585908724) for rationale.

PR #748, opened 2025-01-30
chore: use range.contains in StatusCode methods

DaniPopes:
Small change for readability, alongside some minor documentation touchups for consistency.


seanmonstar (approved):
Thanks!

PR #751, opened 2025-02-16
Added support for title case header name

russellwmy:
I encountered an [issue](https://discuss.ai.google.dev/t/bug-request-header-is-not-properly-handled-for-v1beta-openai-chat-completions/65731) with the OpenAI compatibility API endpoint while using `reqwest`. It turns out the endpoint requires the `host` header to be in title case (`Host`). 

This PR adding a new feature flag to allow developer to use title case header name.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,17 +24,13 @@ edition = "2018"
 rust-version = "1.49.0"
 
 [workspace]
-members = [
-  ".",
-]
-exclude = [
-  "fuzz",
-  "benches"
-]
+members = ["."]
+exclude = ["fuzz", "benches"]
 
 [features]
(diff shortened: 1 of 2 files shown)


seanmonstar:
Hey there! We aren't going to change the implementation here, I'm sorry you went through the work. Opening an issue beforehand might save you work in the future.

To your problem, you can enable title case headers with this option: https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.http1_title_case_headers

russellwmy:
In my case, I am developing a WASI WebAssembly Component, which does not support http1_title_case_headers. That's fine—I’ll just use my current repository for that. Thanks!

PR #756, opened 2025-03-11
fix: validate path bytes are at least utf8

seanmonstar:
cc @joelwurtz

### DIFF
--- a/src/uri/path.rs
+++ b/src/uri/path.rs
@@ -22,6 +22,8 @@ impl PathAndQuery {
         let mut query = NONE;
         let mut fragment = None;
 
+        let mut is_maybe_not_utf8 = false;
+
         // block for iterator borrow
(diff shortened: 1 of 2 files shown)


src/uri/path.rs
```diff
@@ -50,7 +52,12 @@ impl PathAndQuery {
                     0x40..=0x5F |
                     0x61..=0x7A |
                     0x7C |
-                    0x7E..=0xFF => {}
+                    0x7E => {}
+
+                    // potentially utf8, might not, should check
+                    0x7F..=0xFF => {
+                        is_maybe_not_utf8 = true;
```
joelwurtz, line 59: nice idea to avoid checking utf8 in most case, maybe we could do something similar in http-parse, will look at it (since utf8 is not common in path)

PR #758, opened 2025-03-12
Add EARLY_HINTS status code

mdevino:
Hello, I just noticed that the [EARLY_HINTS](https://www.rfc-editor.org/rfc/rfc8297#section-2) status is not added, so raising this PR to add it in case it makes sense.


mdevino:
> Sorry for letting this slip, looks good to me!

No worries. Thanks for the review!

PR #760, opened 2025-03-26
Block leading and trailing whitespace in field values

### LINKED ISSUE #245: Remove trailing whitespaces from header values
As stated [here](https://tools.ietf.org/html/rfc7230#section-3.2.4),
> The field value does not
   include any leading or trailing whitespace: OWS occurring before the
   first non-whitespace octet of the field value or after the last
   non-whitespace octet of the field value ought to be excluded by
   parsers when extracting the field value from a header field.

Here's a test case that should be passing:
```rust
#[test]
fn test_trailing_space() {
    let value = HeaderValue::from_bytes("coucou \t ".as_bytes()).unwrap();
    assert_eq!(value.to_str().unwrap(), "coucou");
}
```

DemiMarie:
This blocks header values with leading and trailing whitespace. This is not needed for HTTP/1.x where the parser already strips such whitespace, but in HTTP/2 and HTTP/3 the parser does not strip such whitespace and so it is up to the `http` crate to reject such headers.

Different HTTP/2 and HTTP/3 libraries treat such values differently, so rejection is the safest option.

Fixes #245

### DIFF
--- a/src/header/value.rs
+++ b/src/header/value.rs
@@ -241,7 +241,14 @@ impl HeaderValue {
         src: T,
         into: F,
     ) -> Result<HeaderValue, InvalidHeaderValue> {
-        for &b in src.as_ref() {
+        let u8_slice = src.as_ref();
+        match u8_slice {
+            [b' ', ..] | [b'\t', ..] | [.., b' '] | [.., b'\t'] => {
+                return Err(InvalidHeaderValue { _priv: () })
+            }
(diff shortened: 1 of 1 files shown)


DemiMarie:
I converted this to draft as it is missing tests.

DemiMarie:
This is an up-to-date version of #256, but without tests and without the more useful error messages.

PR #761, opened 2025-04-09
Make `StatusCode::from_u16` const

coolreader18:
Don't even need to feature-gate it or anything, this works on 1.49.


RemiBardon:
@seanmonstar Tests failed but `master` is failing too (didn’t pass on my machine either)

RemiBardon:
@seanmonstar Thank you for looking into this! 🙏🏻

PR #762, opened 2025-04-17
fix: properly parse schemeless URIs

mina86:
Properly parse URIs such as ‘//example.com/path’ as having
‘example.com’ authority and ‘/path’ path (see RFC 3986
§ 4.2.). Without this change, the code would parse them as
all being path.

This also updates Display representation of schemeless URIs.
If such URI includes the authority, the representation now
includes `//` prefix.

### DIFF
--- a/src/uri/mod.rs
+++ b/src/uri/mod.rs
@@ -296,47 +296,34 @@ impl Uri {
             return Err(TooLong.into());
         }
 
-        match s.len() {
-            0 => {
-                return Err(Empty.into());
-            }
-            1 => match s[0] {
-                b'/' => {
-                    return Ok(Uri {
-                        scheme: Scheme::empty(),
-                        authority: Authority::empty(),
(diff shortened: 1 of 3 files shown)


seanmonstar:
Thanks for the PR! However, this is not a change we'd like to make. The `http::Uri` is modeled after the `request-target`, not after all the behavior in RFC 3986. Such that, a parsed `http::Uri` should be valid to send as a request-target in a request.

A scheme-less target is not allowed.

Note that there are other configurations of `http::Uri` that would not be considered an RFC 3986 URI, since they can also be in authority-form, or just an asterisk.

What you want would require providing a base URL to resolve the scheme-less URI with. That might exist in some other crate.

mina86:
Sounds to me it should be renamed from `Uri` to `RequestTarget` then.

seanmonstar:
Some have suggested that. However, it's not currently planned.

PR #768, opened 2025-05-21
Avoid unnecessary .expect()s for empty HeaderMap

akonradi-signal:
This change removes the Result::expect()/unwrap() calls in the constructors for an empty HeaderMap. These calls were provably not going to fail at runtime but rustc's inliner wasn't smart enough to figure that out: strings analysis of compiled binaries showed that the error message to the expect() still showed up in generated code.

There are no behavioral differences as a result of this change.

### DIFF
--- a/src/header/map.rs
+++ b/src/header/map.rs
@@ -447,12 +447,12 @@ impl HeaderMap {
     /// ```
     #[inline]
     pub fn new() -> Self {
-        Self::new_empty()
+        Self::default()
     }
 }
 
-impl<T> HeaderMap<T> {
-    fn new_empty() -> Self {
+impl<T> Default for HeaderMap<T> {
+    fn default() -> Self {
         HeaderMap {
             mask: 0,
             indices: Box::new([]), // as a ZST, this doesn't actually allocate anything
@@ -461,7 +461,9 @@ impl<T> HeaderMap<T> {
             danger: Danger::Green,
         }
     }
+}
 
+impl<T> HeaderMap<T> {
     /// Create an empty `HeaderMap` with the specified capacity.
     ///
     /// The returned map will allocate internal storage in order to hold about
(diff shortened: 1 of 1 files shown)


src/header/map.rs
```diff
@@ -445,12 +445,23 @@ impl HeaderMap {
...
+    fn new_empty() -> Self {
+        HeaderMap {
+            mask: 0,
+            indices: Box::new([]), // as a ZST, this doesn't actually allocate anything
+            entries: Vec::new(),
+            extra_values: Vec::new(),
+            danger: Danger::Green,
+        }
+    }
+
```
tesaguri, line 464:
```suggestion
        HeaderMap {
            mask: 0,
            indices: Box::new([]), // as a ZST, this doesn't actually allocate anything
            entries: Vec::new(),
            extra_values: Vec::new(),
            danger: Danger::Green,
        }
    }
}

impl<T> HeaderMap<T> {
```

I think it's simpler to make the `new` constructor initialize the empty map instead of adding a new method for it. `Vec::new` is implemented just like that, for example.

akonradi-signal: That was my first inclination but it would mean we can't use `Self::new()` in the `Default` impl because `new()` is only defined for `HeaderMap<HeaderValue>`, not `HeaderMap<T>`. If we want to try to eliminate the extra method I'd suggest instead making `new()` and `try_with_capacity(0)` call `Default::default()` and putting the actual definition in there. Alternatively we can move the definition of `new()` to the more general impl block below but that might break calling code that is relying on `new()` to constrain the generic type.

akonradi-signal: I went ahead and did this in the latest commit. After looking at https://doc.rust-lang.org/cargo/reference/semver.html#fn-generalize-compatible I wouldn't be opposed to just making `new()` more generic since that's considered a non-breaking change even if breaks type inference in consumers. I'm happy to do that here if that's what maintainers prefer.

PR #772, opened 2025-07-02
Raise MSRV to 1.63

Kriskras99:
This PR raises the MSRV to 1.63 as said in #731.

This allows the following features:
 - Rust edition 2021
 - Support for a lot more `const` things (I'm planning a PR that makes a lot of functions `const`)


dxvid-pts:
very cool!

seanmonstar:
Thanks for this! We recently increased the MSRV to 1.57, which so far handles all that we _need_. hyper's policy is to only increase when there's a specific new feature that is needed.

PR #773, opened 2025-07-10
feat: show types in `Extensions` debug output

crepererum:


seanmonstar:
Oh, very clever! I think this could be nice, but I want to check with a few others, just in case (I guess it could result in more noise, if people didn't want it...)

cc @cratelyn @rcoh @LucioFranco @hyperium/triage

rcoh:
I love it. I think we have something similar in the Rust SDK internals for our own property bag types.

cratelyn (commented):
belatedly, i wanted to say that i really appreciate this change. it's a nice quality-of-life improvement to make the contents of request/response extensions legible.

thank you, @crepererum! :pray:

Issue #771, opened 2025-06-30
Clarify and refine header module documentation


Sol-Ell:
# Issue
 
- The [overview section](https://docs.rs/http/latest/http/header/index.html#overview) of the header module is exposing collision resolution technique and internal collection type. While it is a minor issue, it needs to indicate that the collision resolution algorithm may change in the future.
- The [deterministic ordering section](https://docs.rs/http/latest/http/header/index.html#deterministic-ordering) states that iteration order is "roughly" ordered by insertion. However, the documentation of [HeaderMap::iter](https://docs.rs/http/latest/http/header/struct.HeaderMap.html#method.iter) and [HeaderMap::iter_mut](https://docs.rs/http/latest/http/header/struct.HeaderMap.html#method.iter_mut) methods clearly states that __iteration order is arbitrary__, but consistent across platforms for the same crate version.
- The [adaptive hashing section](https://docs.rs/http/latest/http/header/index.html#adaptive-hashing) states that hash of standard header is computed statically (i presume at compile time?), but I wasn't able to find any evidence of this.

# Example solution

## `HeaderMap`

The [`HeaderMap`] type is a specialized multimap structure for storing header names and values.

## Overview

[`HeaderMap`] is designed specifically for efficient manipulation of HTTP headers. It supports multiple values per header name and provides specialized APIs for insertion, retrieval, and iteration.

The internal implementation is optimized for common usage patterns in HTTP, and may change across versions. For example, the current implementation uses Robin Hood hashing to store entries compactly and enable high load factors with good performance. However, the collision resolution strategy and storage mechanism are not part of the public API and may be altered in future releases.

## Iteration Order

Iteration order of headers in a [`HeaderMap`] is deterministic but arbitrary. It does not necessarily reflect insertion order. This consistency is guaranteed only across platforms for the same crate version. Code relying on a specific order of headers should not assume they are returned in the order they were inserted.

See `HeaderMap::iter` and `HeaderMap::iter_mut` for details.

## Adaptive Hashing

[`HeaderMap`] uses an adaptive strategy for hashing to maintain fast lookups while resisting hash collision attacks. The default hash function prioritizes performance. In scenarios where high collision rates are detected—typically indicative of denial-of-service attacks—the implementation switches to a more secure, collision-resistant hash function.

## Limitations

A [`HeaderMap`] can store at most 32,768 entries (header name/value pairs). Attempting to exceed this limit will result in a panic.

[`HeaderMap`]: crate::header::HeaderMap

[...]

PR #774, opened 2025-07-14
Docs: Clarify the `HeaderMap` documentaion

Sol-Ell:
- Add a brief description of `HeaderMap` type in the header module.
- Improve documentation of the `HeaderMap` type.

Note: Changing the iteration order of a `HeaderMap` iterator whose order is unspecified is considered a minor breaking change.

Fixes: #771

### DIFF
--- a/src/header/map.rs
+++ b/src/header/map.rs
@@ -14,11 +14,47 @@ use super::HeaderValue;
 pub use self::as_header_name::AsHeaderName;
 pub use self::into_header_name::IntoHeaderName;
 
-/// A set of HTTP headers
+/// A specialized [multimap](<https://en.wikipedia.org/wiki/Multimap>) for
+/// header names and values.
 ///
-/// `HeaderMap` is a multimap of [`HeaderName`] to values.
+/// # Overview
+///
+/// `HeaderMap` is designed specifically for efficient manipulation of HTTP
+/// headers. It supports multiple values per header name and provides
+/// specialized APIs for insertion, retrieval, and iteration.
+///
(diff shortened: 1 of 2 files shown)


src/header/map.rs
```diff
@@ -14,11 +14,48 @@ use super::HeaderValue;
 pub use self::as_header_name::AsHeaderName;
 pub use self::into_header_name::IntoHeaderName;
 
-/// A set of HTTP headers
+/// The `HeaderMap` type is a specialized
```
seanmonstar, line 17: Keep a vert short sentence to start helps when the type is listed on a module page, such as: https://docs.rs/http/latest/http/header/index.html

src/header/map.rs
```diff
@@ -14,11 +14,48 @@ use super::HeaderValue;
...
+/// part of the public API and may be altered in future releases.
+///
+/// # Iteration order
+///
+/// Unless otherwise specified, the order in which items are returned by
+/// iterators from `HeaderMap` methods is arbitrary; there is no guaranteed
+/// ordering between elements yielded by such an iterator. However, for a given
+/// crate version, the iteration order will be consistent across all platforms.
+/// Changing the iteration order of such an iterator is considered as a [minor
+/// breaking change](<https://doc.rust-lang.org/cargo/reference/semver.html>).
```
seanmonstar, line 42: I wouldn't claim this is a minor breaking change. We strive not to make breaking changes, so claiming this would be means we couldn't change it. And since it's always been unspecified, we can continue to make that be the truth: it could change at any moment, and it cannot be depended on.

PR #777, opened 2025-08-13
Add `as_str` to `Version` type

itsibitzi:
Stringifications of the variants of the `Version` are known at compile time so requiring a `Debug` format is not necessary.

This is nice for me specifically because I want to record the HTTP version of queries using `metrics` crate. This crate allows users to record labels as either a `String` or `&'static str`, and given I didn't want to allocate a `String` with the `Debug` format I had to copy-paste the lookup.

### DIFF
--- a/src/version.rs
+++ b/src/version.rs
@@ -42,6 +42,20 @@ impl Version {
     pub const HTTP_3: Version = Version(Http::H3);
 }
 
+impl Version {
+    /// Returns a `&'static str` representation of the HTTP version.
+    pub fn as_str(&self) -> &'static str {
+        match self.0 {
+            Http::Http09 => "HTTP/0.9",
+            Http::Http10 => "HTTP/1.0",
+            Http::Http11 => "HTTP/1.1",
+            Http::H2 => "HTTP/2.0",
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for the PR!

My feeling is that an `as_str()` doesn't fit semantically. Is there some sort of true string representation of a `Version`? I don't think there is.

The way I would solve this myself in another crate is to just `match` and return some static strings that make sense for my metrics.

itsibitzi:
Yep, I'm doing the match statement right now. :)

I thought it might be useful for others to have access to the mapping currently within debug, but statically. The RFCs standardizing HTTP seem to use the `HTTP/xyz` format so I felt that was a canonical-enough form and would save some folks writing (admittedly a small amount of) boiler plate. Though I do note that the previous `Debug` implementation referred to `HTTP/2` as `HTTP/2.0`, and similar for `HTTP/3`, which is not what the RFC used. This could be corrected either in both or just in a `as_str` version.

The function doesn't need to be called `as_str`, I suppose, but I'm not sure what would be a better name that would be easy for people to find. `as_rfc_name`? `as_name`?

PR #781, opened 2025-10-17
feat: add keep alive header

joaoprocopio:
i'm a huge fan of this library, everything about this library makes creating network services extremely pleasant.

as cited in the [rfc 2616](https://datatracker.ietf.org/doc/html/rfc2616#section-13.5.1) `"keep-alive"` is a _hop-by-hop header_ and shouldn't be forwarded to the upstream.  
in my case i'm building a reverse proxy service and i need to prevent this headers from being cached and sent to the upstream, and it would be great to have this header that is needed to be treated inside the enum.

currently i'm just using this workaround.

```rust
const RFC_2616_HOP_BY_HOP_HEADERS: LazyLock<[HeaderName; 8]> = LazyLock::new(|| {
    [
        header::CONNECTION,
        header::HeaderName::from_static("keep-alive"),
        header::PROXY_AUTHENTICATE,
        header::PROXY_AUTHORIZATION,
        header::TE,
        header::TRAILER,
        header::TRANSFER_ENCODING,
        header::UPGRADE,
    ]
});
```

but this code could be simpler. (pretty irrelevant, but ok)

```rust
const RFC_2616_HOP_BY_HOP_HEADERS: &[HeaderName] = &[
        header::CONNECTION,
        header::KEEP_ALIVE,
        header::PROXY_AUTHENTICATE,
        header::PROXY_AUTHORIZATION,
        header::TE,
        header::TRAILER,
        header::TRANSFER_ENCODING,
        header::UPGRADE,
];
```


joaoprocopio:
hi, @seanmonstar!

I appreciate your work in this library, it's amazing!

do you think that this code could make it into the main branch?

PR #786, opened 2025-10-27
feat(uri): make `Authority/PathAndQuery::from_static` const

### LINKED ISSUE #750: Uri/Authority/PathAndQuery::from_static should be const
Currently using `from_static` to build `Uri`, `Authority`, or `PathAndQuery` types can't be done in a const context. From inspecting what they do under the hood, it seems like the parsing logic would be possible to extract into a `const fn`. Is there some other limitation preventing these methods from being const?

My use case is building `Uri` components at compile time for trivial assembly at runtime.

WaterWhisperer:
## Changes

- `Authority::from_static` and `PathAndQuery::from_static` are now const.
- Both perform compile-time validation equivalent to their runtime counterparts using byte-wise scanning.
- For const panics on invalid input, uses an out-of-bounds indexing trick compatible with MSRV 1.49 .

## Explanation

`Uri::from_static` remains non-const. Two blockers under MSRV 1.49:
1) Const slicing of `&'static str` is unavailable, but Uri parsing requires slicing the input into scheme/authority/path+query.
2) Non-standard schemes use Box<ByteStr>, which requires allocation not allowed in const fn.

## Tests

- cargo test passed.
- Doc tests for const usage included implicitly via examples.

## Related issue

Partially fixes #750

### DIFF
--- a/src/uri/authority.rs
+++ b/src/uri/authority.rs
@@ -8,6 +8,19 @@ use bytes::Bytes;
 use super::{ErrorKind, InvalidUri, Port, URI_CHARS};
 use crate::byte_str::ByteStr;
 
+/// Validation result for authority parsing.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum AuthorityError {
+    Empty,
+    InvalidUriChar,
+    InvalidAuthority,
+    TooManyColons,
+    MismatchedBrackets,
+    InvalidBracketUsage,
+    EmptyAfterAt,
+    InvalidPercent,
+}
+
 /// Represents the authority component of a URI.
 #[derive(Clone)]
 pub struct Authority {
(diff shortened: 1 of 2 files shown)


WaterWhisperer:
Hello @seanmonstar  

This PR has been open for some time, and I'd like to politely request a review (mainly marking `Authority/PathAndQuery::from_static` as const, so it can be used in const contexts). Thanks!

seanmonstar:
Thanks for the PR! I think the objective is worthy, for sure!

I do worry about having parsing be duplicated in two different forms, it's possible for one to change without the other. Is it possible to make this change while keeping only one version of the parsing code?

seanmonstar:
I think that makes sense, can the `const fn validate()` return an enum so we can provide better error messages about why it failed? Otherwise, sounds good to me.

PR #791, opened 2025-11-14
Add Sec-Fetch-Dest header name constant

invliD:
See [fetch spec](https://www.w3.org/TR/fetch-metadata/#sec-fetch-dest-header), which is the defining spec according to the [relevant IANA entry](https://www.iana.org/assignments/http-fields/http-fields.xhtml#field-names).


seanmonstar:
Thanks for the PR!

I'm not sure we can add _every_ header name, the goal was to list the most common. Though, there is #573 to better define this policy.

PR #798, opened 2025-11-21
feat(method): add QUERY method

seanmonstar:
IESG has just confirmed promoting QUERY to Proposed Standard: https://lists.w3.org/Archives/Public/ietf-http-wg/2025OctDec/0208.html

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -407,6 +407,10 @@ impl Request<()> {
     {
         Builder::new().method(Method::TRACE).uri(uri)
     }
+
+    // This is purposefully excluded because of potential conflict with the
+    // URI query.
+    // pub fn query() -> Builder
 }
 
 impl<T> Request<T> {
--- a/src/method.rs
+++ b/src/method.rs
@@ -60,6 +60,7 @@ enum Inner {
     Trace,
     Connect,
     Patch,
+    Query,
     // If the extension is short enough, store it inline
     ExtensionInline(InlineExtension),
     // Otherwise, allocate it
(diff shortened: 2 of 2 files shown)


seanmonstar:
It has just been fully accepted today: https://www.rfc-editor.org/info/rfc10008/

So let's merge 😎

src/request.rs
```diff
@@ -407,6 +407,10 @@ impl Request<()> {
     {
         Builder::new().method(Method::TRACE).uri(uri)
     }
+
+    // This is purposefully excluded because of potential conflict with the
+    // URI query.
+    // pub fn query() -> Builder
```
tgross35, line 413: Bit late here, but could this be added as perhaps `http_query` if there is concern about a conflict?

For what it's worth, I don't find the conflict too worrisome: if there were a method to get the URI query, I'd expect that to be called `uri_query`, and this being a zero-param associated function (rather than a method to get the URI query, or a 1-param function to build a GET with one) is a pretty big clue.

seanmonstar: I don't mean confusion, but that it would prevent any addition of `req.query()` since it's on the same type.

Is that lack of this easy builder a big deal? It's also possible to do `Request::builder().method("QUERY")`.

tgross35: It certainly isn't a big deal aside from consistency, these constructors aren't used directly all that frequently (at least IME).

I think there is still a case for spending the `query` name on this purpose, though; the typing savings from `::builder().method("QUERY")` -> `::query()` are quite a bit more than `.uri().query()` -> `.query()`.

Issue #807, opened 2025-12-18
Make `HeaderValue::set_sensitive` available in const contexts


archer-321:
Hey, I'd like to use `HeaderValue::set_sensitive` when defining header value constants with `HeaderValue::from_static`.
In particular, I'd like to mark an OAuth client credential that's compiled into the application as sensitive.

I fully understand that a value that's compiled into the binary is trivial to dump and shouldn't be considered secure. Moreover, I fully understand that no amount of obfuscation will change this. However, many OAuth implementations (like GitHub or Forgejo) always issue client credentials, regardless of whether the client is public or confidential. This means clients are [forced to use the credential][rfc6749-client-auth], which effectively means the credential has to be embedded into the client.

[rfc6749-client-auth]: https://datatracker.ietf.org/doc/html/rfc6749#section-3.2.1

I'd be willing to contribute a PR for this, but I wanted to open an issue to discuss first, as the idea of storing confidential header values as constants seems controversial. In addition to the `const` keyword, a note in the doc comment to discourage developers from embedding secrets might be warranted.

---

Thank you for creating Hyper and this crate!

seanmonstar:
That seems like fair change. Would you like to submit a PR?

PR #808, opened 2025-12-19
Make `HeaderValue::set_sensitive` available in const contexts

archer-321:
Make `HeaderValue::set_sensitive` `const` to allow applications to embed sensitive header values statically.

Considering embedded client secrets are generally a red flag, add a comment to the method's documentation to urge developers not to consider embedded sensitive values secure. `const` use of `set_sensitive` should be limited to use cases where the developer is aware that the embedded secret will be world-readable.

This PR implements hyperium/http#807

### DIFF
--- a/src/header/value.rs
+++ b/src/header/value.rs
@@ -298,6 +298,11 @@ impl HeaderValue {
 
     /// Mark that the header value represents sensitive information.
     ///
+    /// This method is `const` to allow marking constants created with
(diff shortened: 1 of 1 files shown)


seanmonstar:
The MSRV doesn't like the `cosnt { }` expressions. I'd suggest just leaving the test case out, it doesn't do much.

Issue #661, opened 2023-12-29
Any way to convert `HeaderValue` to `Bytes` without copying?


stackinspector:
We have `HeaderValue::from_maybe_shared` to create `HeaderValue` from `Bytes` without copying, so what about the reverse?

tesaguri:
There's no way to do that yet. See #459.

Issue #810, opened 2025-12-26
Implement `std::ops::Index<Range<usize>>` (and friends*) for `HeaderValue`


pyrrho:
<sub>* "And friends," being `RangeFull`, `RangeFrom<usize>`, `RangeInclusive<usize>` etc..</sub>

My motivation for this is to simplify handling list-based header values in a way that avoids copying the HeaderValue's data. Consider `If-None-Match`, which may contain multiple comma-separated entity-tags, or `Cookies` which may contain multiple semicolon-separated cookie-pairs. In both cases, it's necessary to operate on individual values within the list, and sub-slicing is a natural choice;

```rust
let inm = HeaderValue::from_static(
    r#"If-None-Match: W/"67ab43", "54ed21", "7892dd""#,
);
assert_eq!(&inm.as_bytes()[15 .. 25], br#"W/"67ab43""#);
assert_eq!(&inm.as_bytes()[27 .. 35], br#""54ed21""#);
assert_eq!(&inm.as_bytes()[37 .. 45], br#""7892dd""#);
```

The above works well if 1. you can limit operations to `&[u8]`s and 2. you can manage the lifetime of the borrowed-from `HeaderValue`. Both of those limitations seem unnecessary given `HeaderValue`'s `Arc`-like memory characteristic, so it's more natural to my mind to allow something like the below,

```rust
let inm = HeaderValue::from_static(
    r#"If-None-Match: W/"67ab43", "54ed21", "7892dd""#,
);
assert_eq!(inm[15 .. 25], HeaderValue::from_static(r#"W/"67ab43""#));
assert_eq!(inm[27 .. 35], HeaderValue::from_static(r#""54ed21""#));
assert_eq!(inm[37 .. 45], HeaderValue::from_static(r#""7892dd""#));
```

This would be a relatively simple change to make using [Bytes::slice](https://docs.rs/bytes/1.11.0/src/bytes/bytes.rs.html#347-409) to do the heavy lifting.

I haven't put together a PR for the above yet because I see two open questions that I feel a maintainer may want to weigh in on.

First is whether or not this should be done at all. Returning a `HeaderValue` from `Index`, rather than a `&HeaderValue` may be surprising in the broader context. It would also implicitly codify the use of `Bytes` — or some other `Arc`-like memory management system — in the type's interface. I don't see either of these as blockers, given `from_maybe_shared` is precedent for both, and `HeaderValue` has been backed by `Bytes` for years now.

[...]

PR #817, opened 2026-01-19
feat: Extend HeaderValue to provide access to/semantics of backing Bytes

pyrrho:
This change adds (or makes public),
- `HeaderValue::from_shared[_unchecked]` which directly accept a `Bytes`
  object as an argument. This should allow for lwss copying when
  constructing `HeaderValue`s from pre-existing `Bytes` buffers.
- `HeaderValue::as_shared` which returns a clone of the `Bytes` object
  that backs the given `HeaderValue`.
- `HeaderValue::into_shared` which is a self-consuming version of the
  above.
- `HeaderValue::[try_]slice` which mirrors the `Bytes::slice` interface,
  and provides copy-free construction of sub-sliced `HeaderValue`s.
- `HeaderName::into_shared` which is a rename of the previously
  `pub(crate)` `.into_bytes` method. Allows access to a `Bytes`
  representation of a HeaderName, though it will usually copy via
  `Bytes::from_static`.
- `Uri::from_shared` which was previously `pub(crate)` and is now
  `pub` with an expanded doc comment.

### DIFF
--- a/src/header/name.rs
+++ b/src/header/name.rs
@@ -1270,7 +1270,24 @@ impl HeaderName {
         }
     }
 
-    pub(super) fn into_bytes(self) -> Bytes {
+    /// Creates a [`Bytes`] object from this `HeaderName`.
+    ///
+    /// This function will _likely_ copy the standard string representation of
+    /// this `HeaderName` into a new `Bytes` object. If this `HeaderName` is a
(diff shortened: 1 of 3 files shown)


pyrrho:
This is an alternative to #811 with a few more bells and whistles.

Fixes #661 and #810. Part of #459.

PR #827, opened 2026-04-03
Rephrase comment

daalfox:
Rephrase the comment on `map` field in `Extensions` to make it easier to understand


seanmonstar (approved):
Thanks :)

PR #833, opened 2026-05-12
fix: clamp Extend size hint so HeaderMap reserve cannot overflow

### LINKED ISSUE #604: HeaderMap::from_iter panics when size is known
In this code, `HeaderMap::from_iter` panics for a set of headers that can fit into a `HeaderMap`. Most likely the size hint should be capped to 24576 to avoid this panic until too many names are appended.
```
use http::{HeaderMap, HeaderName, HeaderValue};

fn main() {
    // If we add 24576 names here, then we cannot append extra values.
    let mut headers = (0..24575)
        .map(|i| {
            (
                HeaderName::from_bytes(format!("H{}", i).as_bytes()).unwrap(),
                HeaderValue::from_static("0"),
            )
        })
        .collect::<HeaderMap>();
    // Append more values to existing name
    let mut i = 1;
    while i < 8195 && headers.append(
        HeaderName::from_bytes(format!("H{}", 1).as_bytes()).unwrap(),
        HeaderValue::from_str(&i.to_string()).unwrap(),
    ) {
        i += 1;
    }

    assert_eq!(headers.len(), 32769);

    // Does not panic
    HeaderMap::from_iter(headers.iter().map(|(n, v)| (n.clone(), v.clone())));

    // Panics
    HeaderMap::from_iter(headers.iter().map(|(n, v)| (n.clone(), v.clone())).collect::<Vec<_>>());
}
```

SAY-5:
Both `Extend` impls pass the iterator's size hint straight to `reserve`, but that hint can exceed the number of names the table is able to index (e.g. when the iterator yields many appended values for one name), causing a `MaxSizeReached` panic for input that the map can actually hold. This clamps the reservation to the largest value `reserve` will accept.

Closes #604

### DIFF
--- a/src/header/map.rs
+++ b/src/header/map.rs
@@ -2133,11 +2133,15 @@ impl<T> Extend<(Option<HeaderName>, T)> for HeaderMap<T> {
         // Reserve the entire hint lower bound if the map is empty.
         // Otherwise reserve half the hint (rounded up), so the map
         // will only resize twice in the worst case.
-        let reserve = if self.is_empty() {
+        let hint = if self.is_empty() {
             iter.size_hint().0
         } else {
             (iter.size_hint().0 + 1) / 2
         };
+
+        // Clamp the hint so an over-estimate cannot overflow `reserve`.
+        let max_reserve = usable_capacity(MAX_SIZE).saturating_sub(self.entries.len());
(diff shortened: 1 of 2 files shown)


seanmonstar (approved):
Thanks!

PR #840, opened 2026-06-02
feat(method): impl PartialOrd + Ord

mmirate:
This allows Method (and anyone who contains it) to be used as B-tree keys.


seanmonstar (approved):
Thanks!

PR #848, opened 2026-07-01
feat(header): add Reporting-Endpoints header

bardiharborow:
This PR adds support for the `Reporting-Endpoints` header which is defined by [the Reporting API § 3.2](https://w3c.github.io/reporting/#header).

This header [is supported by all modern browsers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Reporting-Endpoints#browser_compatibility). Analysis of [the HTTP Archive dataset](https://har.fyi/) indicates that this is the 73rd most commonly present header.

MDN: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Reporting-Endpoints


seanmonstar:
Thanks for the PR!

I think I wouldn't add this to the standard constants. It's provisional in IANA, the spec is a draft. And it's not that common of a header as you pointed out.

The good news is that `from_static` is now const fn, so anyone can easily do this:

```rust
const REPORTING_ENDPOINTS: HeaderName = HeaderName::from_static("reporting-endpoints");
```

Issue #849, opened 2026-07-04
Implement PartialEq/Eq for InvalidUri?


joshka:
I have a library with its own error type, and one variant wraps `http::uri::InvalidUri`. I would like that error type to derive `PartialEq` and `Eq` so unit tests can compare expected and actual errors directly, but `InvalidUri` currently prevents that.

For example:

```rust
#[derive(Debug, PartialEq, Eq)]
enum Error {
    InvalidUri(http::uri::InvalidUri),
}
```

Today I can work around this with a manual `PartialEq` implementation in my crate, but the only special case is this wrapped `http` error. The main thing I want is boring unit tests that compare errors normally.

Would it be reasonable for `InvalidUri` and `InvalidUriParts` to implement `PartialEq` and `Eq`?

The small patch would be:

```rust
#[derive(Debug, Eq, PartialEq)]
pub struct InvalidUri(ErrorKind);

#[derive(Debug, Eq, PartialEq)]
pub struct InvalidUriParts(InvalidUri);
```

`ErrorKind` already implements `PartialEq` and `Eq`, and this would keep it private. I am not trying to expose or match on the URI error kind.

I also realize this is still a public contract. I saw the previous discussions around error trait impls, especially [#128](https://github.com/hyperium/http/pull/128) and [#204](https://github.com/hyperium/http/pull/204) around `Clone`, and I understand the concern that trait impls can constrain future error internals. `PartialEq`/`Eq` seems like a smaller commitment than `Clone`, but it is still a commitment: users could compare two `InvalidUri` values and observe whether they are considered the same broad error.

My take is that this tradeoff is reasonable for these URI errors because the comparison can stay coarse, `ErrorKind` stays private, and the concrete downstream benefit is allowing library error types to derive equality for tests. I may be underestimating the compatibility cost, though, so I would appreciate maintainer feedback.

I am opening this as an issue even though the possible patch is small, because the useful part is the API decision. If the answer is "no, this is too much public contract for URI errors," that decision would be useful to have recorded for future searches.

One scope question: the same motivation may apply to other opaque conversion errors in this crate, such as `InvalidMethod`, `InvalidStatusCode`, `InvalidHeaderName`, `InvalidHeaderValue`, and `ToStrError`. I am starting with `InvalidUri` because that is the case I ran into, and `InvalidUriParts` is just a wrapper around it.

seanmonstar:
Just curious, what about `assert_matches!()`, or old `assert!(matches!())`.

[...]

PR #850, opened 2026-07-04
feat(uri): derive equality for URI errors

joshka:
## Motivation

I opened [#849](https://github.com/hyperium/http/issues/849) as the durable place for the API tradeoff. This PR is a small concrete version of the change so maintainers can evaluate the actual patch shape.

I have a library error type that wraps `http::uri::InvalidUri`. I would like that error type to derive `PartialEq` and `Eq` for straightforward unit tests, but `InvalidUri` currently prevents that.

## Change

This derives `PartialEq` and `Eq` for `InvalidUri` and `InvalidUriParts`. `ErrorKind` remains private.

The test checks that both public error types implement `Eq` and that repeated `InvalidUri` / `InvalidUriParts` failures can be compared directly.

## Compatibility

I realize this is still a public contract. Prior discussions such as [#128](https://github.com/hyperium/http/pull/128) and [#204](https://github.com/hyperium/http/pull/204) were cautious about trait impls on error types because they can constrain future internals.

My take is that `PartialEq`/`Eq` is a smaller commitment than `Clone`: it does not expose `ErrorKind`, and equality can remain a coarse comparison over the existing broad URI error category. That said, users could observe whether two `InvalidUri` values compare equal, so I am opening this as a draft PR to get maintainer feedback on that tradeoff.

Refs #849.

## Validation

```text
cargo fmt --check
cargo test
```

### DIFF
--- a/src/uri/mod.rs
+++ b/src/uri/mod.rs
@@ -118,11 +118,11 @@ pub struct Parts {
 }
 
 /// An error resulting from a failed attempt to construct a URI.
-#[derive(Debug)]
+#[derive(Debug, Eq, PartialEq)]
 pub struct InvalidUri(ErrorKind);
 
 /// An error resulting from a failed attempt to construct a URI.
-#[derive(Debug)]
+#[derive(Debug, Eq, PartialEq)]
 pub struct InvalidUriParts(InvalidUri);
 
 #[derive(Debug, Eq, PartialEq)]
--- a/src/uri/tests.rs
+++ b/src/uri/tests.rs
@@ -441,6 +441,28 @@ fn test_uri_parse_error() {
 }
 
 #[test]
+fn test_invalid_uri_errors_are_comparable() {
+    fn assert_eq_impl<T: Eq>() {}
+
+    fn invalid_uri_parts() -> super::InvalidUriParts {
+        let mut parts = super::Parts::default();
+        parts.authority = Some("example.com".parse().unwrap());
+        parts.path_and_query = Some("/".parse().unwrap());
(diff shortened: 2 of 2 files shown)


joshka:
Thanks for considering this. I looked further into the concern that implementing equality could constrain future internal changes.

One relevant standard-library precedent is `TryReserveError`. It originally derived `PartialEq` and `Eq`, and Rust later deliberately changed it into an opaque struct because the private error details were still unsettled ([rust-lang/rust#87408](https://github.com/rust-lang/rust/pull/87408)). The stabilization proposal explicitly left open both adding support for future allocator details and reducing the error to a unit struct, while the public type was stabilized with its equality implementations intact ([rust-lang/rust#87993](https://github.com/rust-lang/rust/pull/87993)).

I did not find an explicit discussion of equality in those reviews, so I would not claim that this settles the question for `InvalidUri`. It does show that std has stabilized equality for an opaque error whose private representation was deliberately left open to substantial change. Together with `Utf8Error` implementing equality despite carrying positional information, that makes me less concerned that these possible future details are necessarily blocking.

My practical motivation remains that `InvalidUri` is a leaf error from a foundational crate. Including it in a downstream error enum prevents that enum from deriving `PartialEq`, moving the implementation work into each downstream crate. My take is that this compositional benefit outweighs the limited compatibility constraint here, particularly if equality is defined without exposing the private kind or making future diagnostic details significant.

That said, the downstream workaround is reasonable, and I understand the preference not to add the contract. I’m happy to leave this undone and close the PR.

Issue #839, opened 2026-05-26
Regression in 1.4.1: `path_and_query` rejects authority-form and empty inputs


zrbecker:
## Summary

PR #826, shipped in 1.4.1, tightened the validator inside `PathAndQuery`'s constructors so that inputs not starting with `/`, `?`, or `#` are rejected, and empty strings are rejected separately. This is a behavior change for inputs that have been accepted since 1.0.

## Reproduction

```rust
http::Uri::builder().path_and_query("127.0.0.1:8080").build()
// 1.4.0: Ok(127.0.0.1:8080)
// 1.4.1: Err(http::Error(InvalidUri(PathDoesNotStartWithSlash)))

http::uri::PathAndQuery::try_from("")
// 1.4.0: Ok(/)
// 1.4.1: Err(InvalidUri(Empty))
```

## Impact

The most visible breakage is in `hyperium/hyper` v1.9.0, whose CONNECT tests (`client_connect_method` and `client_connect_method_with_absolute_uri`) now fail. That, in turn, is what's surfacing the regression in tokio's CI ([example run](https://github.com/tokio-rs/tokio/actions/runs/26429148604)), since tokio runs hyper's test suite as a downstream integration check. Any other crate that constructs CONNECT request-targets through `Uri::builder().path_and_query(...)` is likely to be affected as well.

## Discussion

Hyper's use of `path_and_query` to hold a CONNECT authority-form (`host:port`) is technically a stretch: the documentation only shows `/foo?bar`-style examples. There are cleaner ways to produce the same `Uri` shape today (`Uri::builder().authority("host:port")` or `"host:port".parse::<Uri>()`, both of which put the bytes in the authority slot), but hyper has been using `path_and_query` since 1.0, and downstream code depends on that behavior whether the docs sanction it or not. Other crates may have followed the same pattern; hyper is just the most visible example.

The empty-string case is harder to defend on principle. RFC 3986's `path-abempty` production explicitly permits an empty path component after an authority, so `PathAndQuery::try_from("")` returning `Ok` is consistent with the spec.

[...]

PR #853, opened 2026-07-14
fix(uri): allow empty paths in uri::Builder

seanmonstar:
See #839 

What do you think, @robjtede?


robjtede:
yea i think it makes sense for a builder which doesn't support creating authority form URIs

Issue #712, opened 2024-08-06
RFC: Add `parts_mut() -> &mut Parts` to allow mutate parts directly


Xuanwo:
Hello, I plan to propose an RFC to add `parts_mut() -> &mut Parts` to enable direct mutation of `Parts`. We already have `body_mut`, `headers_mut`, etc., but we lack a function that allows us to simultaneously update headers, methods, and queries.

This change will will allos us to:

```diff
- let (mut parts, body) = req.into_parts();
- do_my_work(&mut parts);
- let req = Request::from_parts(parts, body);
+ do_my_work(req.parts_mut())
```

seanmonstar:
This was brought up before: https://github.com/hyperium/http/issues/372#issuecomment-563548370 and the comments afterwards.

Issue #372, opened 2019-12-10
Allow getting just Parts from a Builder


jameysharp:
Over in non-binary/http-cache-semantics#2, we're using `request::Parts` and `response::Parts` rather than `Request` or `Response` because we never care about the bodies. But the builders are still the easiest way to construct those.

So it'd be nice to be able to take a builder and say  `.parts().unwrap()` instead of `.body(()).unwrap().into_parts().0`. Our callers will likely have full requests and responses so they won't care, this is just nice for situations like our unit tests where we really only need to build the non-body parts.

I think this means adding new methods to the builders along the lines of `pub fn parts(self) -> Result<Parts<T>> { self.inner }`.

Thoughts?

By the way, thanks for changing the builders to by-value instead of by-reference; that turned out to simplify our code by a lot.

seanmonstar:
Is there a particular reason to use the `Parts` in the public API? I'd think that if someone had some `Request<SomeBody>`, and wanted to check the cache, they'd need to deconstruct it, instead of just passing `&req`.

jameysharp:
I was thinking about opening a separate issue for that. Seems to me that `Request` and `Response` should have `parts` and `parts_mut` methods to match the existing `body` and `body_mut` methods?

Or just make both fields of those structs public, at which point basically none of the methods on those types are strictly necessary. I'm not suggesting removing anything, but the existing methods become purely a matter of convenience rather than a necessity.

I'd just prefer not to expose an API that forces the caller to provide information that I don't want and will never use.

jameysharp:
Hmm. I guess if all my API entry-points have unconstrained type-parameters for any `Request` and `Response` arguments then that's a type-level proof that any type of body is fine, and I guess that's good enough from an API design perspective.

Is the Rust compiler smart enough that if I'm passing in a `&Request<T>` for an unconstrained `T` then it doesn't need to monomorphize a separate copy of the code for every `T`? Given that struct layout is unspecified, I'm not sure how it could be; that optimization would only be safe if you're sure that the non-`T` fields are always at the same offset regardless of what `T` is. I'd be sad if having three different type parameters on different arguments led to an exponential explosion of copies of what'd turn out to be identical code in practice. By contrast, passing `&Parts` obviously doesn't involve monomorphization.

[...]

Issue #56, opened 2017-07-14
Should the `Head` type be public?


seanmonstar:
There is currently `request::Head` and `response::Head` exposed publicly. **Should this be part of the public API, or should be the be made private?** This was discussed a little in #22, and postponed to get things working. I'd like to at least be sure that the decision we make here has been reasoned about.

PR #868, opened 2026-09-03
feat(request,response): add borrowed parts accessors

howardjohn:
- [x] The description, docs, and comments (words meant for humans) are written by a human, not by an LLM. (https://seanmonstar.com/micro/20260729-i-want-your-own-words/)

This adds a smaller helper to expose `&Parts`. Today you can only do `into_parts()` but requires ownership.

My main motivation is I have various helpers that require `&Parts`. Today this is called by something that owns req and does into_parts(); but now I have a `&Request` so I cannot use it. Fixing this requires duplicating the same function for `Request` and `Parts`

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -453,6 +453,27 @@ impl<T> Request<T> {
         Request { head: parts, body }
     }
 
+    /// Returns a reference to the request's component parts.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use http::*;
+    /// let request = Request::builder()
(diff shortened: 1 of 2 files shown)


seanmonstar:
This has come up a few times, let me find the references... But in short, the `Parts` type works for `into` because however we store the fields in memory normally, we can always convert them into `Parts`. But exposing a reference would lock the layout in place.

https://github.com/hyperium/http/issues/712
https://github.com/hyperium/http/issues/372
https://github.com/hyperium/http/issues/56

